Issue #27, opened 2025-05-01
Convert CoreMessage to UIMessage where it makes sense


ianmacartney:
When interacting with AI SDK frontend code, there's a preference for UIMessage. There may also be benefits to passing that into the AI SDK (Mastra mentioned the AI SDK choking on "role: tool" messages).

It's mostly a matter of rearranging messages into a friendlier format, with a few differences:

- The `id` field which is assigned by AI SDK (currently populated for response messages but not for the initial user message).
- The `sources` field which is in our steps model but not on each message.
- The `annotations` field which I'm not sure the origin of (it mentions StreamData so maybe it's an internal signal for marking the start/end/segment of streaming text/objects)

Some references:
https://github.com/vercel/ai/issues/3161#issuecomment-2780627913
https://github.com/mastra-ai/mastra/pull/3230/files

ianmacartney:
If we used UIMessage instead of CoreMessage, the only things we'd miss that I can see are extra metadata like `providerOptions`, `experimental_providerMetadata`, `experimental_content`, and tool result error status.

I suspect we could always go from a UIMessage to a CoreMessage, so it might be worth having that be the primitive value, then steps would only differentiate by having:

- finishReason
- usage
- warnings
- request & response http metadata (headers, ids, timestamp, modelId)

ianmacartney:
Strategy for now:
- keep saving core message
- save other fields on the high-level message
- have a function to convert them to UIMessage
- consider in the future passing UIMessage to the AI SDK internally instead of core messages

Issue #20, opened 2025-04-12
Add a `useChat`-equivalent react hook


ianmacartney:
Getting the http-streamed text back along with the messages, whether pending or successful.
Keep track of the thread (also can create a thread if id not passed in), associate with the authenticated user server-side (convex auth or otherwise)

elazdi-al:
Final touch: Integrating it with BetterAuth would be the cherry on top.

ianmacartney:
FYI if you haven't seen, I just added a Discord channel to discuss /
announce agent things:
https://discord.com/channels/1019350475847499849/1369043426561233018

ianmacartney:
The direction will be more option 1, though it won't require using Workflow (which should be stable so let me know if there's something specific you're waiting on. Pause/Resume is in progress already).
It'll use convex queries on the read side, though streaming will still leverage httpAction similar to the PersistentTextStreaming component. I have an example putting this together [here](https://github.com/ianmacartney/ai-agent-chat) that will likely be the inspiration for the useChat component (how to associate a streaming response with a query response).

Some open questions are around the type of the response data - how to include extra data (I don't like the relatively un-typed method of attaching data the AI SDK uses).

Another is whether composing a few explicit things would be better than trying to pack it into a single hook. e.g.:

```ts
const messagesAndOtherData = useQuery(api.messages.list, { threadId, extraArgs });
//const sendMessageWithOtherData = useMutation(api.messages.send);
const [messages, sendMessage, metadata] = useMessagesWithStreaming({
  messages: messagesAndOtherData.messages, // of type UIMessage
  url: http://myapp.convex.site/foo/text/stream, // defaults to convex.site/streamText
  args: (message) => ({ ...message, extraSendArgs }); // optional, o.w. sendMessage expects string/UIMessage/CoreMessage
});
```

KenjiPcx:
useChat is honestly more of a familiarity thing, but here is what I like about it
- Easy helper functions eg: append message, regenerate message etc
- UiMessage is a familiar way to render the messages but more importantly is their data streaming protocol, https://ai-sdk.dev/docs/ai-sdk-ui/streaming-data, I like that I can write data to a dataWriter and it automatically combines itself to the uiMessage, this is really useful for streaming update messages when a workflow is run as a tool? it also allows stuff like artifacts in the vercel chatbot demo, im not sure how the same could be done with convex chat, maybe with like an artifactStreamId field in the message and when we render the artifact

[...]

Issue #1, opened 2025-04-08
Add usage tracking: per-user and globally


ianmacartney:
It makes sense, since it has all the data, to automatically track usage and segment by users.

Some open questions:
- Should it allow arbitrary date range querying? Is bucketing by day ok?
- Should it do global tracking in addition to per-user?
- What are vacuuming requirements?

apostolisCodpal:
I would suggest a simple approach for the initial implementation and depending on community feedback you can decide on more "out-of-the-box" features.

- Bucketing by day covers most scenarios (most apps would need this feature to charge by usage, and usually it is monthly).
- No global tracking. Developers can setup aggregations if they need.
- No vacuuming requirements. Make it opt-in. Developers are responsible to vacuum.

```typescript
const agent = new Agent(components.agent, {
  thread: openai.chat("gpt-4o-mini"),
  textEmbedding: openai.embedding("text-embedding-3-small"),
  instructions: "You are a helpful assistant.",
  trackingOptions: {
    usage: {
      enabled: true,
    },
  },
});
```

batuhanbilginn:
I would like to have custom keys for tracking per project or per account.

ghandic:
This would be huge - its one thing that is blocking me at the moment because I cant access the component tables, so without forking the component I cant actually pull these useage stats unless I make a duplicated table in my app schema.

ianmacartney:
What are the query APIs that folks want? This is what I have in mind:

- Usage is bucketed by hour (that way usage can be calculated for configurable timezones)
- You can query for it by userId or threadId
- You can use a cron to aggregate usage for previous hours and associate it with more custom data.

I'm a bit unsure how much custom tagging could be supported - I'd want to index it, so fetching by an arbitrary number of tags would be tough, but passing that data through wouldn't be too bad, it's just not clear how it'd get fetched on the other side (paginate over all the data in a cron?).

I also wonder if a dedicated Usage component would make sense that this could plug into - if you wanted to store usage from calls not using this framework e.g.

ghandic:
Having an index for modelid/provider would be good. Similar to t3 chat, there's use cases where you allow access to cheaper models at a higher limit than the pricey ones.

ianmacartney:
Yeah I guess there's a few things here:
- Having some identifier that isn't necessarily userId that represents usage attribution (could be user/project/team/organization etc). Having just one of these sounds fine for now - it's "who will pay for this" for usage-based apps.
[...]

PR #31, opened 2025-05-06
Surface more fields in messages table

ianmacartney:
Let's start moving towards a world where the steps tables doesn't need to exist, and we can get all the UIMessage fields from the messages table ( #27 ), to power things like `useChat` ( #20 ).

This adds a bunch of fields to messages, as well as changing the interface to some save methods / component api to add steps. As a result, this is a breaking change for those relying on the current manual saving behavior.

Also makes usage more accessible through the message query API ( #1 )

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,8 +289,12 @@ const messages = await ctx.runQuery(
 ```
 
 ```ts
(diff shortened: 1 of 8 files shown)


ianmacartney:
available in `@convex-dev/agent@0.0.14-alpha.0`

PR #32, opened 2025-05-06
Expose usageHandler for usage tracking

### LINKED ISSUE #1: Add usage tracking: per-user and globally

ianmacartney:
Enables usage tracking (not via a dedicated component yet, but via whatever mechanism you want)

Fixes #1 

Still worth doing:

- [ ] Update README with syntax & basic usage
- [ ] Provide example using aggregate
- [ ] Outline how you could do an hourly cron to get usage data from the messages table instead of the callback

Open Q: Is a Usage component worth it? Is there a general API that's a lot better than bespoke aggregates / tables?

### DIFF
--- a/example/convex/_generated/api.d.ts
+++ b/example/convex/_generated/api.d.ts
@@ -180,11 +180,11 @@ export declare const components: {
                 };
             model?: string;
             provider?: string;
-            providerMetadata?: Record<string, any>;
+            providerMetadata?: Record<string, Record<string, any>>;
             reasoning?: string;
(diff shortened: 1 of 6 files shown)


ianmacartney:
in `@convex-dev/agent@0.0.14-alpha.0` to validate

PR #44, opened 2025-05-12
Fix#38: Ensures proper pairing of tool messages

### LINKED ISSUE #38: Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'.
"convex": "^1.23.0"
"@convex-dev/agent": "0.0.15-alpha.2"

When using `includeToolCalls: true` in `contextOptions`, tool calls are not correctly setup.

```
data: {
    error: {
      message: "Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'.",
      type: 'invalid_request_error',
      param: 'messages.[8].role',
      code: null
    }
  },
```

This is an interesting topic where I have a few questions. Let's take the following example:
```ts
const agentInstance = new Agent(components.agent, {
    name: "Agent",
    instructions,
    maxSteps: 10,
    tools,
    chat: openai.languageModel("gpt-4.1-mini"),
    textEmbedding: openai.embedding("text-embedding-3-small"),
    contextOptions: {
      recentMessages: 10,
      includeToolCalls: true,
      searchOtherThreads: false,
      searchOptions: {
        limit: 3,
        messageRange: { before: 2, after: 2 },
        textSearch: false,
        vectorSearch: true,
      },
    },
  });
```
I declared that I want the 10 most recent messages kept in my content.
1. Do tools count towards the 10 messages?
2. I declared `includeToolCalls` inside my `contextOptions`, not my `searchOptions`. Are old tool calls retrieved by text/vector search?

I believe the answers should be:
1. No - or configurable. I know that tool calls can eat up a lot (if not most) of context in many cases but the bottom line is that they are not messages.
2. No - Only recent history tool calls should be present. If RAG on old tool calls would be a thing, it should be defined under `searchOptions`. Personally I can't think of a use-case for this, unless it is for the model to see old examples of tool calls in cases where tools are very complex.

apostolisCodpal:
Addresses an issue where tool messages were not being properly paired in context.

- Modifies the order in which thread messages are retrieved, changing it to ascending.
- If the topmost message after `recentMessages` cutoff is an orphaned tool response, drops the message.

Fixes #38 

----

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -454,7 +454,7 @@ export class Agent<AgentTools extends ToolSet> {
             cursor: null,
           },
           parentMessageId: args.parentMessageId,
-          order: "asc",
+          order: "desc",
           statuses: ["success"],
         }
       );
@@ -468,15 +468,7 @@ export class Agent<AgentTools extends ToolSet> {
 
     // When including tool calls, drop any orphaned tool messages at the start
     if (opts.includeToolCalls) {
-      while (sortedDocs.length > 0 && sortedDocs[0].tool) {
-        const firstDoc = sortedDocs[0];
-        const nextDoc = sortedDocs[1];
-
-        // Valid pair: first is assistant with tool-call, second is tool with result
-        if (firstDoc.message?.role === 'assistant' && nextDoc?.message?.role === 'tool') {
-          break;
-        }
-        // Otherwise this is an orphaned call or result -> drop it
+      if (sortedDocs.length > 0 && sortedDocs[0].tool && sortedDocs[0].message?.role === "tool") {
         sortedDocs.shift();
       }
     }
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -454,17 +454,35 @@ export class Agent<AgentTools extends ToolSet> {
             cursor: null,
           },
           parentMessageId: args.parentMessageId,
-          order: "desc",
+          order: "asc",
           statuses: ["success"],
         }
       );
       contextMessages.push(...page.filter((m) => !included?.has(m._id)));
     }
-    return contextMessages
-      .sort((a, b) =>
-        a.order === b.order ? a.stepOrder - b.stepOrder : a.order - b.order
-      )
-      .map((m) => deserializeMessage(m.message!));
+
+    // Sort the raw MessageDocs by order and stepOrder
+    const sortedDocs = contextMessages.sort((a, b) =>
+      a.order === b.order ? a.stepOrder - b.stepOrder : a.order - b.order
+    );
+
+    // When including tool calls, drop any orphaned tool messages at the start
+    if (opts.includeToolCalls) {
+      while (sortedDocs.length > 0 && sortedDocs[0].tool) {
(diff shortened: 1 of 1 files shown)


src/client/index.ts
```diff
@@ -454,17 +454,35 @@ export class Agent<AgentTools extends ToolSet> {
             cursor: null,
           },
           parentMessageId: args.parentMessageId,
-          order: "desc",
+          order: "asc",
```
ianmacartney: A problem here is that this will now fetch the first N messages instead of the latest. Let's keep this "desc" and then keep the sorting below.

src/client/index.ts
```diff
@@ -454,17 +454,35 @@ export class Agent<AgentTools extends ToolSet> {
...
           statuses: ["success"],
         }
       );
       contextMessages.push(...page.filter((m) => !included?.has(m._id)));
     }
-    return contextMessages
-      .sort((a, b) =>
-        a.order === b.order ? a.stepOrder - b.stepOrder : a.order - b.order
```
ianmacartney, line 465: This should have been sorting them correctly already - I think the only bug was the orphaned messages

src/client/index.ts
```diff
@@ -454,17 +454,35 @@ export class Agent<AgentTools extends ToolSet> {
...
-      .map((m) => deserializeMessage(m.message!));
+
+    // Sort the raw MessageDocs by order and stepOrder
+    const sortedDocs = contextMessages.sort((a, b) =>
+      a.order === b.order ? a.stepOrder - b.stepOrder : a.order - b.order
+    );
+
+    // When including tool calls, drop any orphaned tool messages at the start
+    if (opts.includeToolCalls) {
+      while (sortedDocs.length > 0 && sortedDocs[0].tool) {
```
ianmacartney: maybe `&& sortedDocs[0]!.message?.role === "tool")`? The orphaned case is just when it starts with a tool response, ya?

ianmacartney (changes requested):
let's keep "desc", but o.w. looking good

src/client/index.ts
apostolisCodpal: I didn't change the sorting.

Ale4224: tool call and tool response have the same order and step order, so they get the original desc order

ianmacartney:
Let’s reverse the list after we get it back in descending order OrdOrdering
by steps and step order is more of a safeguard against differentOrdering by
steps and step order is more of a safeguard against different agents
writingOrdering by steps and step order is more of a safeguard against
different agents writing to the same thread, if I remember correctly

ianmacartney:
More generally the step order should probably be incremented between the to
call and result. It’s a bummer to not have that More generally the step
order should probably be incremented between the to call and result. It’s a
bummer to not have that ordering out of the box.

ianmacartney:
Beyond dangling tool responses, if every step has the response and the call
reversed then it’s going to be trouble throughout

ianmacartney:
I made the reversing change, thanks again for catching & cutting the PR!

PR #58, opened 2025-06-12
feat: add localhost file inlining for LLM requests

julionav:
## Summary
- Add automatic conversion of localhost URLs to base64 data URIs when sending messages to LLMs
- Solves the issue where LLMs cannot access files hosted on localhost during local development

## Rationale
When sending images/files to OpenAI and other providers that support URLs, the ai-sdk sends the URL of the file directly. This works fine in production, but fails in local development mode because:
- The URLs point to localhost (e.g., `http://localhost:3000/file.png`)
- LLM providers cannot access localhost URLs from their servers
- This results in the LLM being unable to process the images/files

This PR solves this by automatically detecting localhost URLs and converting them to base64 data URIs before sending to the LLM.

## Changes
- Added `inlineFilesInLocalDevelopment` config option to Agent constructor (defaults to true)
- Implemented `_processMessagesForLocalhost()` method to process messages before sending to LLM
- Added `_isLocalhostUrl()` helper to detect localhost URLs (supports localhost, 127.0.0.1, ::1, 0.0.0.0)
- Added `_convertUrlToBase64()` helper to fetch and convert files to base64 data URIs
- Integrated processing into `_saveMessagesAndFetchContext()` workflow

## Test plan
- [ ] Test with localhost image URLs in user messages
- [ ] Test with localhost file URLs in user messages
- [ ] Verify non-localhost URLs are not processed
- [ ] Test with `inlineFilesInLocalDevelopment: false` to disable feature
- [ ] Verify error handling when localhost files cannot be fetched

## Pending Actions
- [ ] Write unit tests for `_processMessagesForLocalhost()` method
- [ ] Write unit tests for `_isLocalhostUrl()` helper
- [ ] Write unit tests for `_convertUrlToBase64()` helper
- [ ] Add integration tests with actual LLM calls
- [ ] Manual testing with local Convex instance
- [ ] Test with various file types (images, PDFs, etc.)
- [ ] Test performance impact with large files
- [ ] Update documentation to mention this feature

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -2,12 +2,16 @@ import type { EmbeddingModelV1, LanguageModelV1 } from "@ai-sdk/provider";
 import type {
   CoreMessage,
   DeepPartial,
+  FilePart,
   GenerateObjectResult,
   GenerateTextResult,
+  ImagePart,
   StepResult,
   StreamObjectResult,
   StreamTextResult,
+  TextPart,
   ToolSet,
+  UserContent,
 } from "ai";
 import { generateObject, generateText, streamObject, streamText } from "ai";
 import { assert } from "convex-helpers";
@@ -176,6 +180,13 @@ export class Agent<AgentTools extends ToolSet> {
        * log the raw request body or response headers to a table, or logs.
        */
       rawRequestResponseHandler?: RawRequestResponseHandler;
+      /**
+       * When file or image URLs point to localhost, this will inline them
+       * by converting to base64. This is necessary because LLMs cannot
+       * access localhost URLs.
+       * Defaults to true.
+       */
+      inlineFilesInLocalDevelopment?: boolean;
     }
   ) {}
 
@@ -1495,6 +1506,12 @@ export class Agent<AgentTools extends ToolSet> {
       order = saved.messages.at(-1)?.order;
       stepOrder = saved.messages.at(-1)?.stepOrder;
     }
+    // Process messages to inline localhost files
+    const processedMessages = await this._processMessagesForLocalhost([
+      ...contextMessages.map((m) => deserializeMessage(m.message!)),
+      ...messages,
+    ]);
+
     const { prompt: _, model, ...rest } = args;
     return {
       args: {
@@ -1502,10 +1519,7 @@ export class Agent<AgentTools extends ToolSet> {
         maxRetries: args.maxRetries ?? this.options.maxRetries,
         model: model ?? this.options.chat,
         system: args.system ?? this.options.instructions,
-        messages: [
-          ...contextMessages.map((m) => deserializeMessage(m.message!)),
-          ...messages,
-        ],
+        messages: processedMessages,
       } as T & { model: LanguageModelV1 },
       userId,
       messageId,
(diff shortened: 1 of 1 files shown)


julionav:
@ianmacartney any idea how can I check if the current convex is a local development? That would help to run all this logic only in local mode and clean things a bit

src/client/index.ts
```diff
@@ -176,6 +179,13 @@ export class Agent<AgentTools extends ToolSet> {
        * log the raw request body or response headers to a table, or logs.
        */
       rawRequestResponseHandler?: RawRequestResponseHandler;
+      /**
+       * When file or image URLs point to localhost, this will inline them
+       * by converting to base64. This is necessary because LLMs cannot
+       * access localhost URLs.
+       * Defaults to true.
+       */
+      inlineFilesInLocalDevelopment?: boolean;
```
ianmacartney: what's the usecase for passing false here? a localhost LLM for testing? I wonder if we can get away with not offering the option to start and adding it if folks need it, to have one less thing for someone to see / think they need to make a decision on

src/client/index.ts
```diff
@@ -1495,17 +1505,28 @@ export class Agent<AgentTools extends ToolSet> {
...
+
+    let processedMessages = [
+      ...contextMessages.map((m) => deserializeMessage(m.message!)),
+      ...messages,
+    ];
+
+    // Process messages to inline localhost files (if not, file urls pointing to localhost will be send to LLM provviders)
+    if (
+      this.options.inlineFilesInLocalDevelopment !== false &&
+      process.env.CONVEX_CLOUD_URL === "http://127.0.0.1:3210"
```
ianmacartney: If there's other existing backends when you provision, we may pick another port like `3212`. Maybe anything on `127.0.0.1`? The nice thing is that folks are unlikely to accidentally end up with a prod issue because of this - they'd run into the issue well before it bites them and we could narrow this as necessary.

src/client/index.ts
```diff
@@ -1608,6 +1629,85 @@ export class Agent<AgentTools extends ToolSet> {
...
+    return (
+      url.hostname === "localhost" ||
+      url.hostname === "127.0.0.1" ||
+      url.hostname === "::1" ||
+      url.hostname === "0.0.0.0"
+    );
+  }
+
+  /**
+   * Convert a localhost URL to base64 data URI
```
ianmacartney: technically this isn't true, since the data URI would likely look more like `image/png;base64,...` ? And this isn't base64 encoded?

ianmacartney (commented):
This is awesome, thanks for doing this. just a few nits

src/client/index.ts
julionav: Not a big use case -- just playing it safe. Updating now.

src/client/index.ts
julionav: Got it -- sounds good 👍

src/client/index.ts
julionav: yep, this is outdated -- fixing 👍

julionav:
Should be good for another pass @ianmacartney

src/client/index.ts
```diff
@@ -1608,6 +1619,85 @@ export class Agent<AgentTools extends ToolSet> {
...
+   * able to access localhost URLs.
+   */
+  private async _inlineMessagesFiles(
+    messages: CoreMessage[]
+  ): Promise<CoreMessage[]> {
+    // Process each message to convert localhost URLs to base64
+    return Promise.all(
+      messages.map(async (message): Promise<CoreMessage> => {
+        if (
+          message.role !== "user" ||
```
ianmacartney, line 1634: we might want to drop this check to account for assistant-generated images. but it maybe makes the discriminated union more annoying below..

src/client/index.ts
```diff
@@ -1608,6 +1619,85 @@ export class Agent<AgentTools extends ToolSet> {
...
+      url.hostname === "0.0.0.0"
+    );
+  }
+
+  /**
+   * Download a file from a URL
+   */
+  private async _downloadFile(url: URL): Promise<ArrayBuffer> {
+    // Fetch the file
+    const response = await fetch(url);
```
ianmacartney, line 1693: if this wasn't localhost it'd be worth adding some cache here since you might run multiple generations over the same messages, but fine for now

ianmacartney:
nice work!

ianmacartney:
out in ***@***.******@***.***`

PR #78, opened 2025-07-10
Fixed initial prompt message pending bug

weswinder:
When sending prompts they would hang in the pending state since they aren't tool calls. This fixes that.

----


ianmacartney:
Fixed in the latest version

PR #89, opened 2025-07-26
Fix small error in docs/threads.mdx

reeceyang:
----


docs/threads.mdx
```diff
@@ -31,7 +31,7 @@ You may also pass in metadata to set on the thread:
 
 ```ts
 const userId = await getAuthUserId(ctx);
-const { thread } = await agent.createThread(ctx, {
+const { threadId } = await agent.createThread(ctx, {
```
ianmacartney, line 34: In an action context, it will actually return both of these, but for consistency with the above example you're right this makes more sense

PR #104, opened 2025-08-04
Update a single `tool-[toolName]` part in `toUIMessages`

defrex:
This PR updates `toUIMessage` to better match the UI Message format from AI SDK v5.

It does this in two ways.

Firstly by updating a single tool-call part with `type: tool-${toolName}`, rather than a `type: 'tool-call'` for the input and `type: tool-${toolName}` for the output.

Secondarily, I noticed the `output` was in the wrong format. It looked like `{ type: "json", value: { [actual output] }`. This differs from how it's handled in ai sdk v5, which uses the JSON directly on the `output` field similar to `input`.

This PR addresses both issues with accompanying unit tests.

**Note:** I've addressed the `ai-sdk-v5` branch with this PR, since that seems to be the WIO support branch for AI SDK v5.

----

### DIFF
--- a/src/react/toUIMessages.test.ts
+++ b/src/react/toUIMessages.test.ts
@@ -243,10 +243,9 @@ describe("toUIMessages", () => {
     expect(uiMessages).toHaveLength(1);
     expect(uiMessages[0].role).toBe("assistant");
     expect(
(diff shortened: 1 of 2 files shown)


ianmacartney:
it should reflect in the pkg-pr-new version shortly

PR #139, opened 2025-08-28
Fix message bubble UI for agents + newline wrapping

jamieday:
There's a UI bug with the MessageItem component:
1. Message bubble UI doesn't show for agents
2. There's no wrapping of newlines for both agents & users

Before fix:   
image

After fix:  
image

----


ianmacartney (approved):
thanks!

PR #148, opened 2025-09-04
Exposes the triggers instance from the agent package

alifarooq9:
exposes the triggers instance from the agent package, enabling devs to register triggers directly in their own apps on agent component tables


ianmacartney:
If we were to capture this as a new feature - maybe having an "onSaveMessages" mutation that's called in the same transaction as messages being saved would be good? You could do things transactionally with your own data when messages are saved / updated. This could be a big win for folks who want to generate associated data when tool calls happen, e.g. for human-in-the-loop usecases. 
Would that achieve what you're looking for @alifarooq9 ?

alifarooq9:
> If we were to capture this as a new feature - maybe having an "onSaveMessages" mutation that's called in the same transaction as messages being saved would be good? You could do things transactionally with your own data when messages are saved / updated. This could be a big win for folks who want to generate associated data when tool calls happen, e.g. for human-in-the-loop usecases. Would that achieve what you're looking for @alifarooq9 ?

Yes, this would be a better approach.

PR #159, opened 2025-09-16
Add core memory and memory management features

mantrakp04:
- Introduced core memory functionality with create, update, append, prepend, and remove operations.
- Implemented memory management with add, modify, remove, and list capabilities.
- Enhanced search functionality for memories using embeddings.
- Updated API definitions to include new core memory and memory methods.
- Added tests for core memory and memory functionalities to ensure reliability.
- Included a new utility script for renaming components and managing case variations.

This commit lays the groundwork for improved memory handling in the application.

----

### DIFF
--- a/src/component/coreMemories.test.ts
+++ b/src/component/coreMemories.test.ts
@@ -70,7 +70,7 @@ describe("coreMemories", () => {
     });
     const got = await t.query(api.coreMemories.get, { userId: "u3" });
     expect(got?._id).toBe(doc?._id);
-    expect(got?.persona).toBe("AB");
+    expect(got?.persona).toBe("\nA\nB");
   });
 
   test("replace and remove operate correctly", async () => {
@@ -93,15 +93,12 @@ describe("coreMemories", () => {
     let got = await t.query(api.coreMemories.get, { userId: "u4" });
     expect(got?.human).toBe("heXXo");
 
-    // remove the "XX" segment
+    // remove the entire document
     await t.mutation(api.coreMemories.remove, {
       userId: "u4",
-      field: "human",
-      index: 2,
-      length: 2,
     });
     got = await t.query(api.coreMemories.get, { userId: "u4" });
-    expect(got?.human).toBe("heo");
+    expect(got).toBe(null);
   });
 
   test("getOrCreate prevents duplicates per userId", async () => {


rename.mjs
```diff
@@ -0,0 +1,259 @@
...
+    console.error("❌ Component name is required!");
+    process.exit(1);
+  }
+
+  // Prompt for npm package name
+  const npmPackageName = await new Promise((resolve) => {
+    rl.question(
+      `Enter your npm package name [@convex-dev/${toKebabCase(componentName)}]: `,
+      (answer) => {
+        resolve(answer.trim() || `@convex-dev/${toKebabCase(componentName)}`);
```
ianmacartney: If we're adding this back in, it likely shouldn't assume it's going to be a `@convex-dev/*` package

src/client/index.ts
```diff
@@ -1670,4 +1670,4 @@ function errorToString(error: unknown): string {
     return error.message;
   }
   return String(error);
-}
+}
```
ianmacartney, line 1681:
```suggestion
}
```

src/client/search.ts
```diff
@@ -524,31 +546,52 @@ export async function fetchContextWithPrompt(
...
         inputMessages,
         inputPrompt,
         existingResponses,
+        coreMemory: coreMemoryMessages,
         userId,
         threadId,
       })
     : [
         ...search,
+        ...coreMemoryMessages,
```
ianmacartney, line 579: what do you think about before/after search messages? was this intentional?
reminder: update any docs on contextHandler that mention the default ordering

src/component/coreMemories.ts
```diff
@@ -0,0 +1,138 @@
...
+  },
+  returns: v.null(),
+  handler: async (ctx, args) => {
+    const doc = await ctx.db
+      .query("coreMemories")
+      .withIndex("userId", (q) => q.eq("userId", args.userId))
+      .first();
+    assert(doc, `Core memory for user ${args.userId} not found`);
+    const base = (doc[args.field] as string | undefined) ?? "";
+    await ctx.db.patch(doc._id, { [args.field]: args.text + base });
```
ianmacartney: maybe separate with newline?

src/component/messages.ts
```diff
@@ -60,7 +60,6 @@ export async function deleteMessage(
 
 export const deleteByIds = mutation({
   args: { messageIds: v.array(v.id("messages")) },
-  returns: v.array(v.id("messages")),
```
ianmacartney: why?

src/component/schema.ts
```diff
@@ -137,15 +137,14 @@ export const schema = defineSchema({
...
   }).index("streamId_start_end", ["streamId", "start", "end"]),
 
-  memories: defineTable({
-    threadId: v.optional(v.id("threads")),
+  coreMemories: defineTable({
     userId: v.optional(v.string()),
-    memory: v.string(),
-    embeddingId: v.optional(vVectorId),
+    persona: v.optional(v.string()),
+    human: v.optional(v.string()),
```
ianmacartney: drop a comment here about what these terms mean. not clear to me

src/component/vector/index.ts
```diff
@@ -149,18 +149,28 @@ export function searchVectors(
...
+      }
+      if (args.threadId) {
+        return q.eq("model_table_threadId", [
+          args.model,
+          args.table,
+          args.threadId,
+        ]);
+      }
+      // Fallback to model+table only – this broadens search but avoids undefined values
+      return q.eq("model_table_threadId", [args.model, args.table, ""]);
```
ianmacartney: I don't think this will match anything, since "" isn't a valid threadId

test-assets/mornin_cat.jpeg
ianmacartney, line 1: y?

ianmacartney (commented):
high level question: does this mean folks will have ~8 more tool calls on every agent LLM call (4 for each of short & long memory)? Or maybe one tool call with multiple options/params?

test-assets/mornin_cat.jpeg
mantrakp04: i had to change git origins and had to check if it was syncing to the fork instead of the private unforked repo real quick, i didn't know what to change so i added this.

src/client/search.ts
mantrakp04: i am still testing out various memory solutions, graph based, rag, notepad, namespaces, etc.

I initially thought to have this with the rag based memories but then saw better results with the message based overall. Will seperate core memories from the search.

src/component/messages.ts
mantrakp04: red squiggly lines

src/component/schema.ts
mantrakp04: human is for user preferences and persona is for agent guidance, thinking of adding a namespace but don't want multiple notepad memories.

Issue #33, opened 2025-05-07
Allow branching message history


ianmacartney:
Allow specifying a `parentMessageId` on thread creation. Behavior:
- `recentMessages` will pull from the parent thread if there aren't enough messages in the new thread. If the parent thread doesn't have enough, it will go to its parent, etc.
- search will **not** search in any parent thread, only in the current thread. This is mostly a side-effect of how the search indexes are implemented right now. To fetch from a specific parent, you could always fetch context messages from the desired thread, then feed those into the generateText call as a list of messages directly.. 

The workaround for this currently is to copy the old messages over manually

ianmacartney:
I'm starting to think that doing a full copy would be a nice convenience function. But not a new capability. 
And then anything else can be done in a developer's own tables with custom logic to fetch message history from parent threads in whatever way makes sense.. eg you could make a table with a threadId and parent message(s) and do fetchMessageContext on the parent messages

PR #161, opened 2025-09-17
Ian/clone thread

ianmacartney:
In the vein of #33 , allows cloning messages from a thread into a new thread up to some point, adding at some offset.

It avoids copying if messages exist at that new order/stepOrder

----


ianmacartney:
@KevinGrajeda not a good one. I cut it for folks to try out & give feedback but never got confirmation that it solved the problem so it just sat. Have you tried it? Does it work for you?

KevinGrajeda:
@ianmacartney confirming that works for me, I’m using this version in production because my hand‑rolled approach was breaking on larger threads, but this one has been working flawlessly

PR #176, opened 2025-10-20
Fix tool calls not being persisted when tools error  

a-rebets:
**Problem:**

When a tool execution throws an error, the assistant message containing the tool-call is not saved to the database. Only the tool-result message with the error is persisted, breaking the message chain.

**Root Cause:**

In `src/mapping.ts`, the `serializeNewMessagesInStep` function uses `step.toolResults.length` to determine which messages to save:

```ts
const messages: MessageWithMetadata[] = await Promise.all(
  (step.toolResults.length > 0
    ? step.response.messages.slice(-2)  // assistant + tool messages
    : step.content.length
      ? step.response.messages.slice(-1)  // Only last message
      : [{ role: "assistant" as const, content: [] }]
  ).map(...)
);
```

When a tool fails:
- `step.toolResults` is empty (no successful execution result)
- `step.response.messages` contains both the assistant message with tool-call AND the tool message with error
- The code incorrectly takes only the last message, missing the assistant message

What gets saved to the database (example):
```js
[
  {
    content: "list the datasets",
    role: "user",
  },
  // MISSING: assistant message with tool-call
  {
    content: [
      {
        output: {
          type: "error-text",
          value: "Your request couldn't be completed. Try again later.",
        },
        toolCallId: "call_somehash",
        toolName: "datasets_listDatasetsTool",
        type: "tool-result",
      },
    ],
    role: "tool",
  },
]
```

After all of this we see warnings like:
```
[WARN] 'Tool result without preceding tool call.. adding anyways'
```

**Proposed Fix:**

Check if there are tool messages in `step.response.messages` instead of relying solely on the length

----

### DIFF
--- a/src/mapping.ts
+++ b/src/mapping.ts
@@ -204,7 +204,7 @@ export async function serializeNewMessagesInStep<TOOLS extends ToolSet>(
   } satisfies Omit<MessageWithMetadata, "message" | "text" | "fileIds">;
   const toolFields = { sources: step.sources };
   const messages: MessageWithMetadata[] = await Promise.all(
-    (step.toolResults.length > 0
+    (step.finishReason === "tool-calls"
       ? step.response.messages.slice(-2)
       : step.content.length
         ? step.response.messages.slice(-1)


src/mapping.ts
```diff
@@ -204,7 +204,7 @@ export async function serializeNewMessagesInStep<TOOLS extends ToolSet>(
   } satisfies Omit<MessageWithMetadata, "message" | "text" | "fileIds">;
   const toolFields = { sources: step.sources };
   const messages: MessageWithMetadata[] = await Promise.all(
-    (step.toolResults.length > 0
+    (step.finishReason === "tool-calls"
```
ianmacartney, line 207: in the case of error, is the finishReason `tool-calls`?
My understanding of `tool-calls` finishReason was that it was the signal that there were tool calls that need response, e.g. when you have a tool with no `execute` function, it will finish with `tool-calls` but there will only be one response message - the assistant response.

ianmacartney:
released in `@convex-dev/agent@0.2.12` - let me know if you see any issues

PR #186, opened 2025-11-19
strip out old streams when they are no longer present

### LINKED ISSUE #185: Duplicate messages with stuck streaming state after tool approval (saveStreamDeltas)
## Summary

When using `@convex-dev/agent` with `useUIMessages` hook and `saveStreamDeltas: true`, approving or denying a tool call results in duplicate assistant messages appearing in the UI:
1. **Stream delta message**: Partial/incomplete text, stuck with `status: "streaming"` even after stream completes, ID has "stream" prefix
2. **MessageDoc**: Complete text with `status: "success"`, regular message ID

Closing and reopening the UI component makes the duplicate disappear and shows only the correct complete message.

## Environment

- `@convex-dev/agent`: 0.3.0
- `ai`: 5.0.51
- React Native mobile app
- Using `useUIMessages` hook with `stream: true`
- Backend uses `thread.streamText()` with `saveStreamDeltas: true`

## Reproduction Steps

1. Set up agent with human-in-the-loop tool approval flow
2. User sends message that triggers a tool call (e.g., "Schedule a meeting")
3. Assistant responds with tool-call, approval UI shows
4. User approves or denies the tool call
5. Backend saves tool result message to thread using `chatAgent.saveMessage()`
6. Backend calls `thread.streamText()` with `contextOptions: { recentMessages: 20 }` and `saveStreamDeltas: true`
7. Stream completes successfully on backend
8. Frontend shows **two assistant messages** instead of one

## Expected Behavior

After tool approval, there should be one assistant message responding to the tool result with:
- Full text content
- `status: "success"` (or appropriate final status)
- Single message ID

## Actual Behavior

Two assistant messages appear:
1. **Stream delta (incomplete/stuck)**:
   - Text is incomplete or partial
   - `status: "streaming"` (never transitions to "finished")
   - ID has "stream" prefix

2. **MessageDoc (complete)**:
   - Full text content
   - `status: "success"`
   - Regular message ID

### Backend Logs Show Correct Behavior

```
[STREAM_FINISH] Stream completed: { threadId, responseMessages: 1, finishReason: 'stop' }
[STREAM_DONE] Stream consumed

[TOOL_RESULT] All messages AFTER stream: [
  { id: 'ks75jy3hx8gasg180f2mawxxrs7vdn5g', role: 'assistant', status: 'success', order: 0, stepOrder: 3 },
  { id: 'ks71dp83bxp9wtby4xtpdw6wz17vdyfb', role: 'tool', status: 'success', order: 0, stepOrder: 2 },
  { id: 'ks7av80wa18h4h7q9zebr3rgp17vdzpe', role: 'assistant', status: 'success', order: 0, stepOrder: 1 },
  { id: 'ks71qdg4fkvtv8zn24zhx8cxvx7vc2d5', role: 'user', status: 'success', order: 0, stepOrder: 0 }
]
```

✓ Backend has only ONE assistant message at stepOrder=3
✓ Status is correctly "success"
✓ Stream consumed successfully

### Frontend Receives Duplicate

`useUIMessages` returns two messages:
- One with "stream" prefix ID - partial/stuck at "streaming"
- One with regular message ID - complete with "success"

### Workaround That Proves It's Client-Side State Issue

**Closing and reopening the UI component fixes the issue immediately.**

This proves:
- Backend data is correct (otherwise reopening wouldn't fix it)
- Problem is in `useUIMessages` internal state management
- Stream deltas are persisting in component state after stream finishes

## Suspected Root Cause

When `saveStreamDeltas: true` is used:
1. Stream deltas are saved to `streamingMessages` table during streaming
2. `useUIMessages` merges these deltas with MessageDoc objects
3. After stream finishes, the stream delta state doesn't update from `"streaming"` to `"finished"`
4. Frontend continues showing both:
   - The stream delta (status: "streaming", partial text)
   - The MessageDoc (status: "success", complete text)
5. Remounting the component reinitializes `useUIMessages` state, fetching fresh data and correctly showing only the MessageDoc

### Backend Tool Result Flow

```typescript
// 1. Build and save tool result message
const toolMessage = await buildToolResultMessage(ctx, input, approvalTools, spanTracker);
await chatAgent.saveMessage(ctx, { threadId, message: toolMessage });

// 2. Start new stream (no messages parameter, let context fetch from thread)
const result = await thread.streamText({
  model: "openai/gpt-4.1",
  system: systemPrompt,
  tools,
  // ... callbacks
}, {
  saveStreamDeltas: true,
  contextOptions: {
    recentMessages: 20,
  },
});

await result.consumeStream();
```

### Frontend Query

```typescript
// domains/chat/api.chat.ts
export const listThreadMessages = authorizedQuery({
  handler: async (ctx, { threadId, paginationOpts, streamArgs }) => {
    const paginated = await listUIMessages(ctx, components.agent, {
      threadId,
      paginationOpts,
    });

    const streams = await chatAgent.syncStreams(ctx, {
      threadId,
      streamArgs,
    });

    return { ...paginated, streams };
  },
});

// Frontend
const uiMessagesResult = useUIMessages(
  api.app.chat.listThreadMessages,
  { threadId },
  {
    initialNumItems: 50,
    stream: true,
  },
);
```

## Temporary Workarounds Tried

1. ❌ Filtering finished streams on backend
2. ❌ Aborting old streams before new stream
3. ❌ Updating message status before streaming
4. ✓ **Disabling saveStreamDeltas**: Fixes the issue but loses real-time streaming feature

### LINKED ISSUE #187: Performance: updateFromUIMessageChunks assert eagerly JSON.stringify’s large messages
I’ve been debugging some pretty bad frontend performance in an app using `useThreadMessages` with streaming enabled. In some cases the UI was dropping to 2 FPS 😭

I know we have a fair amount of re‑rendering going on, but even accounting for that, things felt way slower than expected. So I recorded a performance trace in Chrome and found this:

- `updateFromUIMessageChunks` in `deltas.ts` is taking a big chunk of the main thread time 

Image

- Inside that function, this `assert` line seems to be the hotspot :

Image

https://github.com/get-convex/agent/blob/b48f51c76172ad73ce7a117515932896c2dcd799/src/deltas.ts#L80-L83

it is surprising, because this is just an equality check. But my theory is that, regardless of whether the equality check passes, the `JSON.stringify` calls in the template string are always evaluated eagerly.

Since `updateFromUIMessageChunks` is called for every delta while streaming, this means we’re doing a full `JSON.stringify` of the whole message on every new delta.

```txt
useThreadMessages({ stream: true })
  ↓
useStreamingThreadMessages
  ↓
useStreamingUIMessages
  ↓
useDeltaStreams (fetches stream deltas)
  ↓
useEffect in useStreamingUIMessages (on new deltas)
  ↓
updateFromUIMessageChunks
```
this happens with both useThreadMessages and useUIMessages

### Possible fix

If my understanding is correct, this should be a relatively easy performance win by making the error message lazy, so `JSON.stringify` only runs on failure.

Happy to provide more details if helpful.

ianmacartney:
Cleans up streaming messages from the list when no new deltas arrive but a stream is no longer present.

Fixes #185 
Fixes #187 

----

## Summary by CodeRabbit

* **Refactor**
  * Improved stream message processing to avoid work when no new updates are present, reduce redundant state changes, and minimize churn when streams are removed—resulting in more efficient and stable stream handling.

### DIFF
--- a/src/react/useDeltaStreams.ts
+++ b/src/react/useDeltaStreams.ts
@@ -103,36 +103,51 @@ export function useDeltaStreams<
     | undefined;
 
   const newDeltas = cursorQuery?.streams.deltas;
-  if (newDeltas?.length && streamMessages) {
+  if (streamMessages) {
     const newDeltasByStreamId = new Map<string, StreamDelta[]>();
-    for (const delta of newDeltas) {
-      const oldCursor = cursors[delta.streamId];
-      if (oldCursor && delta.start < oldCursor) continue;
-      const existing = newDeltasByStreamId.get(delta.streamId);
-      if (existing) {
-        const previousEnd = existing.at(-1)!.end;
-        assert(
-          previousEnd === delta.start,
-          `Gap found in deltas for ${delta.streamId} jumping to ${delta.start} from ${previousEnd}`,
-        );
-        existing.push(delta);
-      } else {
-        assert(
-          !oldCursor || oldCursor === delta.start,
-          `Gap found - first delta after ${oldCursor} is ${delta.start} for stream ${delta.streamId}`,
-        );
-        newDeltasByStreamId.set(delta.streamId, [delta]);
+    if (newDeltas?.length) {
+      for (const delta of newDeltas) {
+        const oldCursor = cursors[delta.streamId];
+        if (oldCursor && delta.start < oldCursor) continue;
+        const existing = newDeltasByStreamId.get(delta.streamId);
+        if (existing) {
+          const previousEnd = existing.at(-1)!.end;
+          assert(
+            previousEnd === delta.start,
+            `Gap found in deltas for ${delta.streamId} jumping to ${delta.start} from ${previousEnd}`,
+          );
+          existing.push(delta);
+        } else {
+          assert(
+            !oldCursor || oldCursor === delta.start,
+            `Gap found - first delta after ${oldCursor} is ${delta.start} for stream ${delta.streamId}`,
+          );
+          newDeltasByStreamId.set(delta.streamId, [delta]);
+        }
       }
     }
     const newCursors: Record<string, number> = {};
+    let cursorsChanged = false;
     for (const { streamId } of streamMessages) {
       const cursor =
         newDeltasByStreamId.get(streamId)?.at(-1)?.end ?? cursors[streamId];
       if (cursor !== undefined) {
         newCursors[streamId] = cursor;
+        if (cursors[streamId] !== cursor) {
+          cursorsChanged = true;
+        }
       }
     }
-    setCursors(newCursors);
+    // Also check if any streams were removed from cursors
+    for (const streamId in cursors) {
+      if (!(streamId in newCursors)) {
+        cursorsChanged = true;
+        break;
+      }
+    }
+    if (cursorsChanged) {
+      setCursors(newCursors);
+    }
 
     // we defensively create a new object so object identity matches contents
     state.deltaStreams = streamMessages.map((streamMessage) => {


robelest:
Both issues here are already resolved, so closing as superseded.

**#185** was fixed by 54d6f78, which added the length-zero cleanup at `useDeltaStreams.ts:88-92`. Verified on main: after the approval flow, `streams.list` returns zero streams (default `["streaming"]`), so `state.deltaStreams` is cleared and there's no streamed copy left to duplicate.

**#187** is closed and unrelated.

Reopen if I've missed something.

PR #189, opened 2025-11-25
Ian/workflow examples

ianmacartney:
----

### DIFF
--- a/example/tsconfig.json
+++ b/example/tsconfig.json
@@ -6,7 +6,7 @@
     },
     "baseUrl": ".",
     "lib": ["DOM", "DOM.Iterable", "ESNext"],
-    "skipLibCheck": false,
+    "skipLibCheck": true,
     "allowSyntheticDefaultImports": true,
     "strict": true,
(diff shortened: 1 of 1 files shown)


KenjiPcx:
Hey, sorry was out on vacation too for the past week, ive added some more examples at #204, think i figured out an implementation of the in workflow pooling, also my code maybe written in an unoptimized way, do check it out

I think one thing im uncertain about is how we define context within a workflow, because its code, are we able to use in memory variables within a workflow function? does that persist?

Or for context variables that we want to expose in the frontend, whats the right approach to do it? I've been creating a separate table in convex but there could be opportunity for some syntax sugar

ianmacartney:
Catching up here

> are we able to use in memory variables within a workflow function? does that persist?

Yes you can use in-memory variables. They'll be deterministically brought back to the same values on each invocation.

KenjiPcx:
Okay no worries, am I doing the workflow wrong? I have a PR merging towards your branch

PR #194, opened 2025-12-02
feat: split playground API for Node.js-only LLM providers

Minoo7:
Adds two new functions to allow separation of V8-safe queries/mutations from Node.js-only actions in the playground API:

- **`definePlaygroundQueries`** - V8-safe queries and mutations
- **`definePlaygroundActions`** - Node.js actions for text generation

## Problem

When using LLM providers that require Node.js modules (e.g., **Vertex AI** with `google-auth-library`, `http`, `https`), the current `definePlaygroundAPI` causes bundler errors because:

1. Queries/mutations **MUST run in V8 runtime** (Convex restriction)
2. Actions **CAN use Node.js runtime** (with `"use node"` directive)
3. The combined `definePlaygroundAPI` pulls all dependencies (including Node.js-only ones) into the V8 bundle, causing errors like:
   ```
   Could not resolve "http"
   Could not resolve "https" 
   Could not resolve "stream"
   ```

## Solution

Split the API into two separate functions that can be imported independently:

### Usage Example

```typescript
// convex/playground.ts (V8 runtime - no "use node")
import { definePlaygroundQueries } from "@convex-dev/agent";
import { components } from "./_generated/api";

export const { isApiKeyValid, listAgents, listUsers, listThreads, listMessages, createThread } = 
  definePlaygroundQueries(components.agent, {
    agents: [{ 
      name: "MyAgent", 
      instructions: "...", 
      tools: ["tool1", "tool2"] 
    }],
  });
```

```typescript
// convex/playgroundActions.ts (Node.js runtime)
"use node";
import { definePlaygroundActions } from "@convex-dev/agent";
import { components } from "./_generated/api";
import { myVertexAgent } from "./agent";

export const { generateText, fetchPromptContext } = 
  definePlaygroundActions(components.agent, { 
    agents: [myVertexAgent] 
  });
```

## Changes

- Added `definePlaygroundQueries()` - takes static `AgentInfo[]` instead of `Agent[]` to avoid importing Node.js dependencies
- Added `definePlaygroundActions()` - takes `Agent[]` and exports only actions
- Added new types: `AgentInfo`, `PlaygroundQueriesAPI`, `PlaygroundActionsAPI`
- **Preserved `definePlaygroundAPI`** for backward compatibility with providers that work in V8 (e.g., OpenAI via fetch)

### DIFF
--- a/src/client/definePlaygroundAPI.ts
+++ b/src/client/definePlaygroundAPI.ts
@@ -35,12 +35,459 @@ export type PlaygroundAPI = ApiFromModules<{
   playground: ReturnType<typeof definePlaygroundAPI>;
 }>["playground"];
 
+export type PlaygroundQueriesAPI = ApiFromModules<{
+  playground: ReturnType<typeof definePlaygroundQueries>;
+}>["playground"];
+
+export type PlaygroundActionsAPI = ApiFromModules<{
+  playground: ReturnType<typeof definePlaygroundActions>;
+}>["playground"];
+
 export type AgentsFn<DataModel extends GenericDataModel> = (
   ctx: GenericActionCtx<DataModel> | GenericQueryCtx<DataModel>,
   args: { userId: string | undefined; threadId: string | undefined },
 ) => Agent[] | Promise<Agent[]>;
 
-// Playground API definition
+/**
+ * Configuration for agent info displayed in queries when the actual Agent
+ * instance cannot be accessed (due to V8/Node.js runtime separation).
+ */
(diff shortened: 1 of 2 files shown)


ianmacartney:
High level:
1. I don't see where the playground client is being configured to find the actions in a separate place from the queries
2. I want the implementation to be shared for the combined & separate implementations - e.g. to have the unified one call both individually and return both. Too much duplicated code right now , from what I can tell

PR #204, opened 2025-12-27
add more workflow examples

KenjiPcx:
Added some more examples over here
- Actor-Critique Pattern: Implements an iterative self-improvement loop where an "Actor" drafts content and a "Critic" provides scored feedback for automatic revision.
- Council of Agents (Ensemble): Demonstrates parallel "Fan-out" to specialized agents (Analyst, Creative, Pragmatist) followed by a "Fan-in" where a Judge scores or synthesizes the final result.
- Advanced Batching: Showcases three ways to handle high-volume tasks: Fixed-size batches (rate limiting), Worker Pools (efficiency), and Nested Sub-workflows (fault tolerance).
- Deep Agent (Delegation): Uses a multi-turn orchestration pattern where a main agent delegates sub-tasks to sub-workflows and integrates results via completion injection.

### DIFF
--- a/example/convex/workflows/deep_agent.ts
+++ b/example/convex/workflows/deep_agent.ts
@@ -13,7 +13,7 @@ import {
 import { getAuthUserId } from "../utils";
 import { defaultConfig } from "convex/agents/config";
 import { z } from "zod/v4";
-import { GenerateTextResult, TypedToolCall } from "ai";
+import { TypedToolCall } from "ai";
 
 /**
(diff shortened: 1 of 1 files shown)


ianmacartney:
For PR organization, let's:
1. Merge in the #216 branch
2. Fix or remove the workflow examples I added (which currently aren't passing tests)
3. Target the #216 branch as the base.

If you want to have a separate PR just for these additional workflows, that's fine too. But tests need to be passing before Seths reviews / merges anything

PR #210, opened 2026-01-16
Fix stopWhen triggering on tool validation errors

### LINKED ISSUE #172: Agent stopWhen stops even when tool call fails Zod validation
Hey,
When using `stopWhen` with `hasToolCall(...)`, the agent stops execution even when the tool call **fails argument validation** (e.g., Zod returns an error).

In this case, the tool call never actually succeeds, but the agent transitions to a stopped state (`output-available`) instead of retrying or handling the validation error gracefully.

---

**Example setup**

```ts
stopWhen: [
  hasToolCall("generateImage"),
],
```

**Observed behavior**
If the agent calls a tool with invalid arguments (e.g., the Zod validator rejects it), the `stopWhen` condition still triggers, stopping the agent.

Example state:

```
State:
output-available

Type:
tool-generateImage

Arguments:
-- invalid / malformed --

Return Value:
"ArgumentValidationError: Value does not match validator.\nPath: .id\nValue: \"kn...??\"\nValidator: v.id(\"images\")\n\n"
```

**Expected behavior**
The agent should **not** stop when the tool call fails validation.
Instead, it should:

* Recognize the tool call as *unsuccessful*, and
* Retry or continue reasoning to produce a valid tool call.

Only *successful* tool calls (i.e., those that pass validation and run) should trigger `hasToolCall(...)` conditions inside `stopWhen`.

If the tool returns an error, that’s a separate issue; however, the agent should attempt to call the tool again when the initial invocation fails.

sethconvex:
## Problem

When a tool call fails Zod validation (e.g., invalid ID format), the `stopWhen` condition was still being evaluated and could trigger a stop. This prevented the LLM from seeing the error and retrying with corrected arguments.

For example, with `stopWhen: hasToolCall("generateImage")`:

1. LLM calls `generateImage` with invalid ID
2. Tool returns error: `ArgumentValidationError: Value does not match validator`
3. `stopWhen` sees `generateImage` was called → returns `true` → stops
4. LLM never gets a chance to retry with valid arguments

## Solution

Add a check in `willContinue` to skip `stopWhen` evaluation when any tool result has an error:

```typescript
// src/client/utils.ts - willContinue()
export async function willContinue(steps, stopWhen) {
  const step = steps.at(-1)!;
  if (step.finishReason !== "tool-calls") return false;
  if (step.toolCalls.length > step.toolResults.length) return false;
  
  // NEW: If any tool result has an error, continue without evaluating stopWhen.
  // This allows the LLM to retry or handle the error.
  const hasErrorResult = step.toolResults.some(
    (result: { isError?: boolean }) => result.isError === true,
  );
  if (hasErrorResult) return true;
  
  // ... evaluate stopWhen conditions
}
```

Now when a tool fails:
1. LLM calls `generateImage` with invalid ID  
2. Tool returns error with `isError: true`
3. `willContinue` sees error → returns `true` (continue)
4. LLM receives error, retries with valid arguments

Fixes #172

## Test plan
- [x] Test: continues when tool result has error (even if stopWhen would match)
- [x] Test: stops normally when tool succeeds and stopWhen matches
- [x] Test: handles array of stopWhen conditions
- [x] Test: stops when any stopWhen condition returns true

### DIFF
--- a/src/client/utils.ts
+++ b/src/client/utils.ts
@@ -11,6 +11,12 @@ export async function willContinue(
   if (step.finishReason !== "tool-calls") return false;
   // we don't have a tool result, so we'll wait for more
   if (step.toolCalls.length > step.toolResults.length) return false;
+  // If any tool result has an error, continue without evaluating stopWhen.
+  // This allows the LLM to retry or handle the error.
+  const hasErrorResult = step.toolResults.some(
+    (result) => (result as { isError?: boolean }).isError === true,
+  );
+  if (hasErrorResult) return true;
   if (Array.isArray(stopWhen)) {
     return (await Promise.all(stopWhen.map(async (s) => s({ steps })))).every(
       (stop) => !stop,
--- a/src/client/utils.test.ts
+++ b/src/client/utils.test.ts
@@ -1,0 +1,161 @@
+import { describe, it, expect } from "vitest";
+import { willContinue } from "./utils.js";
+import type { StepResult } from "ai";
+
+describe("willContinue", () => {
+  it("should return false when finishReason is not tool-calls", async () => {
+    const steps = [
+      {
+        finishReason: "stop",
+        toolCalls: [],
+        toolResults: [],
+      },
+    ] as unknown as StepResult<any>[];
+
+    const result = await willContinue(steps, undefined);
+    expect(result).toBe(false);
+  });
+
+  it("should return false when waiting for tool results", async () => {
+    const steps = [
+      {
+        finishReason: "tool-calls",
+        toolCalls: [{ toolCallId: "1", toolName: "test", args: {} }],
+        toolResults: [],
+      },
+    ] as unknown as StepResult<any>[];
+
+    const result = await willContinue(steps, undefined);
+    expect(result).toBe(false);
+  });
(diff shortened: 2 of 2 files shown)


src/client/utils.ts
```diff
@@ -11,6 +11,12 @@ export async function willContinue(
   if (step.finishReason !== "tool-calls") return false;
   // we don't have a tool result, so we'll wait for more
   if (step.toolCalls.length > step.toolResults.length) return false;
+  // If any tool result has an error, continue without evaluating stopWhen.
+  // This allows the LLM to retry or handle the error.
+  const hasErrorResult = step.toolResults.some(
+    (result) => (result as { isError?: boolean }).isError === true,
+  );
+  if (hasErrorResult) return true;
```
ianmacartney, line 19: I am concerned that the interpretation of `willContinue` implies that we should make another pending message - but generally these failures won't result in another round of execution (or if it does, we should probably update the current call instead of making a new one?

sethconvex: yes, after discussion on the issue #172 upgrading models fixed this, and I don't think this strategy is universally a good choice.

PR #212, opened 2026-01-16
Preserve isError flag for tool error reporting in UIMessages

### LINKED ISSUE #162: The tools not reporting proper errors with tools
Issue:
When the tools error out the part of the UI message shows the state as "output-available" instead of "output-error". 

Reason at least what I was able to find:
The issue is because in the createAssistantUIMessage function, in tool-result case we are checking for tool-call case's output which is not present in the allParts variable. So it goes in the condition and checks for 'error' field in the messages table which is also empty.

I hope the reason help in providing the right solution. I tried adding data to the error field which I think is not the right solution as if there are two tool calls and and one succeeds and other fails then both will be tagged as fail. I think the underlying issue is that tool-call log is not being created. I also noticed that if there are multiple tool calls the a record is created but still the output is not correct.

Image

Image

The first one did error out but is not showing the same. But as we can see in thedashboard a record was created.

Image

Image

When both the tool calls error out the tool-call record is not created. Also when there is an error the args data is also lost for the same reason as the tool-call record is not created.

sethconvex:
## Problem

When a tool call fails (e.g., validation error), the error was not being properly displayed in the UI. Tool results have an `isError` flag, but it was being stripped during message mapping, so the UI showed errors as normal output.

The `isError` flag was lost here:

```typescript
// src/mapping.ts - normalizeToolResult() - BEFORE
return {
  type: part.type,
  output: part.output ?? normalizeToolOutput(...),
  toolCallId: part.toolCallId,
  toolName: part.toolName,
  ...metadata,
  // isError was NOT preserved!
};
```

And in UIMessages, only `message.error` was checked, not `contentPart.isError`:

```typescript
// src/UIMessages.ts - BEFORE
if (message.error) {
  call.state = "output-error";
  call.errorText = message.error;
}
```

## Solution

1. Preserve `isError` flag during mapping:

```typescript
// src/mapping.ts - normalizeToolResult() - AFTER
return {
  type: part.type,
  output: ...,
  toolCallId: part.toolCallId,
  toolName: part.toolName,
  // NEW: Preserve isError flag for error reporting
  ...("isError" in part && part.isError ? { isError: true } : {}),
  ...metadata,
};
```

2. Check both `contentPart.isError` and `message.error` in UIMessages:

```typescript
// src/UIMessages.ts - AFTER  
// Check for error at both the content part level (isError) and message level
const hasError = contentPart.isError || message.error;
const errorText = message.error || (hasError ? String(output) : undefined);

if (hasError) {
  call.state = "output-error";
  call.errorText = errorText;
}
```

Now tool errors display correctly with `output-error` state.

Fixes #162

## Test plan
- [x] Test: shows output-error state when tool result has `isError: true`
- [x] Test: shows output-error when tool result has isError without tool call present

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for tool results with improved error detection, consolidated error sources, and fallback error messaging when details are unavailable.

* **Tests**
  * Added test coverage for tool result error handling scenarios.

### DIFF
--- a/src/UIMessages.ts
+++ b/src/UIMessages.ts
@@ -463,6 +463,12 @@ function createAssistantUIMessage<
             typeof contentPart.output?.type === "string"
               ? contentPart.output.value
               : contentPart.output;
+          // Check for error at both the content part level (isError) and message level
+          // isError may exist on stored tool results but isn't in ToolResultPart type
+          const hasError =
+            (contentPart as { isError?: boolean }).isError || message.error;
+          const errorText =
+            message.error || (hasError ? String(output) : undefined);
           const call = allParts.find(
             (part) =>
               part.type === `tool-${contentPart.toolName}` &&
@@ -470,9 +476,9 @@ function createAssistantUIMessage<
               part.toolCallId === contentPart.toolCallId,
           ) as ToolUIPart | undefined;
           if (call) {
-            if (message.error) {
+            if (hasError) {
               call.state = "output-error";
-              call.errorText = message.error;
+              call.errorText = errorText ?? "Unknown error";
               call.output = output;
             } else {
               call.state = "output-available";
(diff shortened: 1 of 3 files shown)


ianmacartney (approved):

robelest:
Already on main, so closing.

`isError` preservation landed in dcf844c. Main has it in both serialization directions where this PR touched one:

- `src/vercel/mapping.ts:814` (`normalizeToolResult`), identical line including the comment
- `src/vercel/mapping.ts:834` (`deserializeToolResult`)

The `output-error` handling is at `src/vercel/UIMessages.ts:589-592`, using the same `hasError` approach. Thanks for the fix.

PR #213, opened 2026-01-16
Add userId field to UIMessage type

### LINKED ISSUE #188: Support userId in UIMessage for multi-user thread conversations
## Problem

When working with threads that have multiple users participating in the same conversation, the `userId` information is lost when converting `MessageDoc` to `UIMessage` via `toUIMessages()`.

Currently, `MessageDoc` includes a `userId` field (which is properly saved when creating messages), but when `createUserUIMessage()` converts these to `UIMessage` objects, the `userId` is not included in the returned object. This makes it difficult to identify which user sent each message in multi-user threads.

## Current Behavior

Looking at `UIMessages.ts`:

- `fromUIMessages()` accepts `userId` in the metadata and includes it in `MessageDoc` (line 72)
- `toUIMessages()` converts `MessageDoc[]` to `UIMessage[]` but loses the `userId`
- `createUserUIMessage()` (lines 302-353) creates a `UIMessage` but doesn't include `userId` from the source `MessageDoc`

## Use Case

I have threads where multiple users can participate in the same conversation. When displaying messages in the UI, I need to:
1. Show which user sent each message
2. Filter messages by user
3. Implement user-specific features (e.g., notifications, permissions)

## Proposed Solution

Add `userId` to the `UIMessage` type and include it in `createUserUIMessage()`:

export type UIMessage<...> = AIUIMessage<...> & {
  // ... existing fields
  userId?: string; // Add this
};And in `createUserUIMessage()`:t
return {
  // ... existing fields
  userId: message.userId, // Add this
};
## Alternative Approaches

If adding `userId` directly to `UIMessage` isn't the right approach, I'd appreciate guidance on:
- Should `userId` be part of the `metadata` field instead?
- Is there a better pattern for handling multi-user threads?
- Should I query `MessageDoc` directly when I need `userId` information?

## Questions

1. Is adding `userId` to `UIMessage` the recommended approach?
2. Are there any concerns or breaking changes this might introduce?
3. Would this also need to be added to `createSystemUIMessage()` and `createAssistantUIMessage()` for consistency?

Thank you for your consideration!

sethconvex:
## Problem

The `UIMessage` type did not include a `userId` field, making it impossible to identify which user a message belongs to in the UI layer. This information was available in the underlying `MessageDoc` but was not propagated to `UIMessage`.

## Solution

Add `userId` field to the `UIMessage` type and propagate it during message creation:

```typescript
// src/UIMessages.ts - UIMessage type
export type UIMessage<...> = {
  id: string;
  key: string;
  order: number;
  stepOrder: number;
  status: UIStatus;
  agentName?: string;
  userId?: string;  // NEW
  text: string;
  _creationTime: number;
};
```

Propagate in each message type creator:

```typescript
// System messages
function createSystemUIMessage(...) {
  return {
    ...
    userId: message.userId,  // NEW
    ...
  };
}

// User messages  
function createUserUIMessage(...) {
  return {
    ...
    userId: message.userId,  // NEW
    ...
  };
}

// Assistant messages (uses first message's userId)
function createAssistantUIMessage(...) {
  const common = {
    ...
    userId: firstMessage.userId,  // NEW
  };
  ...
}
```

Fixes #188

## Test plan
- [x] Test: preserves userId in user messages
- [x] Test: preserves userId in system messages
- [x] Test: preserves userId in assistant messages
- [x] Test: preserves userId from first message in grouped assistant messages
- [x] Test: handles undefined userId gracefully

## Summary by CodeRabbit

* **New Features**
  * Messages now include an optional userId so system, user, and assistant messages preserve and carry a message-level user identifier across flows.

* **Tests**
  * Added tests to verify userId preservation across roles and grouped assistant messages, including edge cases where userId is undefined.

### DIFF
--- a/src/UIMessages.ts
+++ b/src/UIMessages.ts
@@ -42,6 +42,7 @@ export type UIMessage<
   stepOrder: number;
   status: UIStatus;
   agentName?: string;
+  userId?: string;
   text: string;
   _creationTime: number;
 };
@@ -75,7 +76,8 @@ export function fromUIMessages<METADATA = unknown>(
         "providerOptions",
         "metadata",
       ]),
-      ...omit(uiMessage, ["parts", "role", "key", "text"]),
+      ...omit(uiMessage, ["parts", "role", "key", "text", "userId"]),
+      userId: uiMessage.userId ?? meta.userId,
       status: uiMessage.status === "streaming" ? "pending" : "success",
       streaming: uiMessage.status === "streaming",
       // to override
@@ -288,6 +290,7 @@ function createSystemUIMessage<
     text,
     role: "system",
     agentName: message.agentName,
+    userId: message.userId,
     parts: [{ type: "text", text, ...partCommon } satisfies TextUIPart],
     metadata: message.metadata,
   };
@@ -347,6 +350,7 @@ function createUserUIMessage<
     key: `${message.threadId}-${message.order}-${message.stepOrder}`,
     text,
     role: "user",
+    userId: message.userId,
     parts,
     metadata: message.metadata,
   };
@@ -370,6 +374,7 @@ function createAssistantUIMessage<
     stepOrder: firstMessage.stepOrder,
     key: `${firstMessage.threadId}-${firstMessage.order}-${firstMessage.stepOrder}`,
     agentName: firstMessage.agentName,
+    userId: firstMessage.userId,
   };
 
   // Get status from last message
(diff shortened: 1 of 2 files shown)


ianmacartney (approved):
This works - I figured any metadata returned on MessageDoc would show up, but it's a good point that when it's reduced to a UIMessage some fields get "promoted" to not be per-message but for the whole collection, and userId is something that seems fine to special-case. Doesn't seem like we'll be leaking extra data out, since that data was already present by default

PR #217, opened 2026-01-25
Add auto-continuation for tool approval workflow

sethconvex:
When a tool has needsApproval: true and the user provides an approval
response, generation now automatically continues to execute the approved
tools instead of stopping.

Changes:
- Modified willContinue() in utils.ts to check for approval responses
  in messages when tool calls lack results
- Updated call sites in index.ts and streamText.ts to pass messages
- Added comprehensive tests for approval continuation logic

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Tool approval workflow: explicit approve/deny actions, conditional approvals, and demo chat UI with inline controls.
* **Improvements**
  * Better streaming finish coordination and atomic finalization; improved UI message deduplication and approval-state rendering.
  * Extended usage tracking with detailed token breakdowns; added approval-related message fields and indexes for fast lookup.
* **Tests**
  * Extensive new test suites covering approval workflows and edge cases.
* **Documentation**
  * Migration and upgrade docs updated for v0.6.0 and SDK v6 guidance.


CLAUDE.md
```diff
@@ -87,82 +88,138 @@ Follow the rules in `.cursor/rules/convex_rules.mdc`:
...
+**Compatible sibling packages:**
+- `@convex-dev/rag@^0.7.0` (v0.6.0 has type conflicts with AI SDK v6)
+- `@convex-dev/workflow@^0.3.2`
+
+### Step 2: Detect v5 Patterns
 
-### Step 2: Update tool definitions
+Search for these patterns indicating v5 usage:
+- `createTool({ args:` - should be `inputSchema`
+- `createTool({ handler:` - should be `execute`
```
ianmacartney, line 114: we own `createTool` so these are technically up to us - I guess `inputSchema` is fine since it's zod which might be a bit unexpected to be for `args`, but `handler` takes different arguments than `execute` so I'm nervous we're leading people astray by making them match exactly... I know this is the wrong PR for this, but curious what the intent here was.

example/convex/chat/approval.ts
```diff
@@ -0,0 +1,160 @@
...
+    approvalId: v.string(),
+    reason: v.optional(v.string()),
+  },
+  handler: async (ctx, { threadId, approvalId, reason }) => {
+    const result = await approvalAgent.approveToolCall(ctx, {
+      threadId,
+      approvalId,
+      reason,
+    });
+    await result.consumeStream();
```
ianmacartney, line 114: Are streams default for approving/ denying? Or does it remember the previous setting? Do you have to stream? Just curious

example/convex/usage_tracking/usageHandler.ts
```diff
@@ -37,6 +37,22 @@ export const insertRawUsage = internalMutation({
...
+        })
+      ),
+      outputTokenDetails: v.optional(
+        v.object({
+          textTokens: v.optional(v.number()),
+          reasoningTokens: v.optional(v.number()),
+        })
+      ),
+      // Provider-specific raw usage data (varies by provider)
+      raw: v.optional(v.any()),
```
ianmacartney: is this always an object? Maybe `v.record(v.string(), v.any())` ?

package.json
```diff
@@ -109,7 +110,7 @@
     "chokidar-cli": "3.0.0",
     "class-variance-authority": "0.7.1",
     "clsx": "2.1.1",
-    "convex": "1.29.3",
+    "convex": "^1.31.6",
```
ianmacartney: let's pin exact versions in devDependencies so we know what tests are running against in case a version bump causes issues. If we don't, renovate will open a PR suggesting to do it, not sure if it'll auto-land, so maybe this will happen automatically

src/client/index.ts
```diff
@@ -19,7 +19,7 @@ import type {
 } from "ai";
 import { generateObject, generateText, stepCountIs, streamObject } from "ai";
 
-const MIGRATION_URL = "https://github.com/get-convex/agent/blob/main/MIGRATION.md";
+const MIGRATION_URL = "node_modules/@convex-dev/agent/MIGRATION.md";
```
ianmacartney: technically this could be deep in a pnpm directory.. maybe we could pin to a version's URL via its tag

src/client/index.ts
```diff
@@ -1508,4 +1508,278 @@ export class Agent<
...
+      },
+      skipEmbeddings: true,
+    });
+
+    // Continue generation with forceNewOrder to create a separate message
+    return this.streamText(
+      ctx,
+      { threadId },
+      { promptMessageId: toolResultId, forceNewOrder: true },
+      { saveStreamDeltas: { chunking: "word", throttleMs: 100 } },
```
ianmacartney: We shouldn't assume they want to save deltas or assume the chunking strategy here. I suggest we save the message and return the messageId and have them call streamText with promptMessageId themself to make it explicit. Then they can also decide whether to do it, e.g. if there were multiple tool responses that need to finish before continuing.   
I also will look into what this `forceNewOrder ` parameter is, because I don't think that should be necessary, and fear it may break other assumptions about how messages are combined. e.g. the assumption that only user messages increment the order.

src/client/streamText.ts
```diff
@@ -145,13 +145,24 @@ export async function streamText<
...
+      options?.saveStreamDeltas === true
+    ) {
+      await stream;
+      await result.consumeStream();
+      streamConsumed = true;
+    }
+  } catch (error) {
+    // If an error occurs during streaming (e.g., in onStepFinish callbacks),
+    // make sure to abort the streaming message so it doesn't get stuck
+    if (streamer && !streamConsumed) {
```
ianmacartney: it seems impossible for streamConsumed to be true at this point, based on my understanding of JS awaits.

ianmacartney (changes requested):

zboyles:
> Thanks for your work on this. I hoped the needsApproval flow on createTool were what we needed, but I'm running into an issue with runtime tools though and wondering if I'm missing something. We have ~20 tools passed to streamText at call time (they need per-request workspace context via closures) and 4 destructive tools on the agent definition via createTool with needsApproval.
> 
> The initial streamText call gets all tools via the tools parameter. But after approveToolCall, the continuation only sees this.options.tools (the 4 agent-level ones)....so if the user instructs our agent to "delete X and then create Y" and approves the delete, the agent can't use createWhatever in the continuation ...because it was a runtime tool.
> 
> We worked around it but it's super fragile since we're depending on internal message formats and the __acceptsCtx pattern. Is there a cleaner way to handle this that I'm missing? Or would it make sense for approveToolCall/denyToolCall to accept an optional tools parameter for the continuation?

@kerns 
If you used the pkg-pr-new version, we've decided on some changes that should fix it for you. We're going to split out the approval and the continuation. This means the approval tool call and the message that gets created, optionally with the tool results, will be part of one call, then you'll call your start stream function `streamText` which should let you pass your other tools to the continuation process. These two functions would be called sequentially inside the approve/deny approval handlers.

sethconvex:
## Fix Applied: Atomic Stream Finish

This commit fixes the race condition causing `TypeError: The stream is not in a state that permits enqueue` when using `saveStreamDeltas` with tool calls.

### Root Cause
When tools execute quickly, the stream can close before delta-save mutations complete:
1. Tool executes → `onStepFinish` queues mutation
2. Stream finishes → AI SDK closes stream  
3. Mutation still in flight → crashes when writing to closed stream

### Solution
Restored the atomic finish pattern from commit 87e3657 (which existed on an unmerged branch):
- Defer final step save when streaming enabled
- Save atomically with stream finish in same mutation
- Stream stays open until DB confirms completion

### Verification
- ✅ Typecheck passes
- ✅ No conflicts with approval flow (tool execution happens before final step)
- ✅ API compatible with current codebase
- ✅ All changes from proven working commit 87e3657

The fix is safe - it was developed in parallel but never merged during AI SDK 6.0 upgrade.

eliotgevers:
Any update regarding this change? Will this be merged relatively soon, or will a different solution be pursued to handle parallel tool calls that need approval?

robelest:
Closing as unmergeable, not unwanted. The feature is now tracked in #352, which carries the design decisions from these 22 commits.

The blocker is AI SDK generation. This branch targets `ai@^6.0.35` with `@ai-sdk/provider-utils@^4` and providers at `^3`; main is on `ai@^7`, `provider-utils@^5`, providers at `4.0.x`, and #306 was a breaking migration rather than a bump.

On top of that: 144 commits behind, 22 merge conflicts, several in files that no longer exist at those paths after the `src/` to `src/vercel/` move, and the conflicts land in exactly the files this changes (`createTool.ts`, `messages.ts`, `start.ts`, `streaming.ts`, `validators.ts`). `willContinue` has also been rewritten twice since, by #241 and #263, so the function this modifies is not the one it was written against.

What carried over to #352: save-only approve/deny with deferred execution, atomic stream finish with message save, the atomic idempotency check for the TOCTOU window, indexed O(1) approval lookup, and streaming the continuation into a separate message. The ~1,900 lines of tests here, particularly `approval-bugs.test.ts`, are worth reading before anyone reimplements.

Thanks for the work, and sorry it sat long enough to fall behind a major version.

PR #219, opened 2026-01-28
01 24 add auto continuation for tool approval workflow

zboyles:
This refactors the tool approval process to support all tool result output types - tools returning anything other than a string were failing to continue. The utility function `createToolModelOutput`, from AI SDK, is used to help construct the approval `ToolContent` parts.

The `createTool` parameter types were refactored to resolve a type issue, introduced with the removal of the deprecated types paired with the addition of the warning types, which surfaced when extending `createTool` into a factory function.

**Agent tool execution and approval flow improvements:**

* Refactored the `approveToolCall` method in the `Agent` class to support richer tool result outputs, including structured error handling and output formatting via a new `createToolModelOutput` utility. The method signature was updated to accept separate `threadOpts`, `approvalArgs`, and `streamTextArgs`, and now returns more detailed streaming results. (`src/client/index.ts`, [[1]](diffhunk://#diff-24f1ac19d22913269d99a8a5afa4c13aa7fd2c8b9858529046424841ca6421daL1519-R1589) [[2]](diffhunk://#diff-24f1ac19d22913269d99a8a5afa4c13aa7fd2c8b9858529046424841ca6421daL1554-R1671) [[3]](diffhunk://#diff-24f1ac19d22913269d99a8a5afa4c13aa7fd2c8b9858529046424841ca6421daL1589-R1696)
* Introduced the `createToolModelOutput` helper function in `src/shared.ts` to standardize the formatting of tool outputs and errors for model consumption. (`src/shared.ts`, [src/shared.tsR18-R57](diffhunk://#diff-3cc490302b79321eb38cc902ba2bc669600333b6e8428e4ac75981e3d5262550R18-R57))

**Dependency and import updates:**

* Added new imports for types and utilities required by the new tool result handling, such as `Tool`, `StaticToolResult`, `StaticToolError`, and `ToolContent`. (`src/client/index.ts`, [[1]](diffhunk://#diff-24f1ac19d22913269d99a8a5afa4c13aa7fd2c8b9858529046424841ca6421daR6) [[2]](diffhunk://#diff-24f1ac19d22913269d99a8a5afa4c13aa7fd2c8b9858529046424841ca6421daR15-R20); `src/shared.ts`, [[3]](diffhunk://#diff-3cc490302b79321eb38cc902ba2bc669600333b6e8428e4ac75981e3d5262550R5-R8)
* Updated imports to include the new `createToolModelOutput` utility where needed. (`src/client/index.ts`, [src/client/index.tsL49-R57](diffhunk://#diff-24f1ac19d22913269d99a8a5afa4c13aa7fd2c8b9858529046424841ca6421daL49-R57))


ianmacartney:
What do you think about having it return the messageId and having the user call the streamText / generateText / generateObject immediately after? That way it's a bit more transparent, and exposes all of the options available? I'm not sure saving the extra line is worth it - I think explicit and more verbose here is ok.
I'm also still trying to understand the need for `forceNewOrder`, but will have that discussion elsewhere. 
In my mind it'd look like:
```ts
const { messageId } = agent.approveToolCall(ctx, { threadId }, { approvalId: "cross-agent-approval" });
const result = await agent.streamText(ctx, { threadId}, { promptMessageId: threadId, ... });
```
I think separating them could help keep the approveToolCall more "pure" - you could run that in a workflow step without retries, then do the LLM generation part with retries, e.g.

robelest:
Closing alongside #217, which this branches from. The feature is tracked in #352.

Worth recording that these were never competing implementations: this targets #217's branch rather than `main`, so its diff against main includes all of #217, which is why it looks nearly twice the size.

What this added on top is carried into #352: support for tools returning non-string outputs, which were failing to continue, using AI SDK's `createToolModelOutput` to build the approval `ToolContent` parts. That was a real bug and a real fix.

The blocker is the same as #217's. The branch targets AI SDK v6 and main is on v7 after the breaking migration in #306, with 144 commits of drift and 22 conflicts in the exact files this changes.

Thanks for the work.

PR #257, opened 2026-05-05
fix cross-thread order filtering in textSearch

### LINKED ISSUE #256: searchOtherThreads drops valid cross-thread text search results when target message order is lower
Using `@convex-dev/agent@0.6.1`

I have an agent configured with:

```js
contextOptions: {
  searchOtherThreads: true,
  searchOptions: {
    limit: 10,
    textSearch: true,
  },
}
```

and my contextHandler includes `args.search`.

**Expected:** a new thread should be able to retrieve matching messages from older threads with the same stored `userId`.

**Actual:** direct text search finds the old matching messages, but they get filtered out when the new thread’s current message has a lower order.

## Example:

- **New thread prompt:** `what do you remember about high-ticket coaches`
- **Old thread message:** `tom and jerry are both amazing high-ticket coaches and educators...`

Both threads have the same `userId`.
Raw `text_search` returns the old message.

**But the agent context search returns no cross-thread match.**

It looks like the “only include messages before `targetMessageId`” filter compares order across different threads. That makes sense within one thread, but not across threads where the search results could be used in order 0 or 1 even though in other threads they may have been order 100.

## Files affected

- `src/component/messages.ts`
`textSearch` filters results using `targetMessage.order`.

- `src/client/search.ts`
passes `targetMessageId` into `searchMessages` while also using `searchOtherThreads`.

## Potential fix

When `searchAllMessagesForUserId` is set, only apply the `targetMessageId` order filter to messages from the same `threadId`.

ianmacartney:
Fixes a bug (#256) where `textSearch` with `searchAllMessagesForUserId` incorrectly filtered out cross-thread matches by applying the target message's `order` as a global filter. Since message order sequences are independent per thread, a target message with a low order in a new thread was causing valid matches from older threads (which naturally have higher order values) to be excluded.

The fix scopes the order-based filter to same-thread messages only, both at the database query level and in the post-fetch filter. When searching across threads, the database-level `lte(order)` constraint is skipped entirely, and the post-fetch filter now checks `threadId` before applying the order comparison. Same-thread ordering behavior is preserved — messages after the target within the same thread are still excluded.

Two regression tests are added to cover both cases: cross-thread matches being returned correctly, and same-thread messages after the target still being filtered out.

Fixes #256

### DIFF
--- a/src/component/messages.test.ts
+++ b/src/component/messages.test.ts
@@ -703,7 +703,7 @@ describe("agent", () => {
         ],
       },
     );
-    const targetMessageId = newMessages[0]._id;
+    const targetMessageId = newMessages[0]._id as Id<"messages">;
 
     const results = await t.query(api.messages.textSearch, {
       searchAllMessagesForUserId: userId,
@@ -759,7 +759,7 @@ describe("agent", () => {
 
     const results = await t.query(api.messages.textSearch, {
       searchAllMessagesForUserId: userId,
-      targetMessageId: targetMessages[0]._id,
+      targetMessageId: targetMessages[0]._id as Id<"messages">,
       text: "high-ticket",
       limit: 10,
     });
--- a/src/component/messages.test.ts
+++ b/src/component/messages.test.ts
@@ -652,4 +652,124 @@ describe("agent", () => {
     );
     expect(remainingMessages.page).toHaveLength(1);
   });
+
+  // Regression test for #256: when searching across threads with
+  // searchAllMessagesForUserId, the targetMessage's order should not filter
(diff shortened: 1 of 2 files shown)


ianmacartney:
* **#257** <a href="https://app.graphite.com/github/pr/get-convex/agent/257?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/agent/257?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

robelest (approved):
LGTM

Issue #240, opened 2026-04-02
willContinue doesn't count tool-error as completed output (AI SDK v6)


impish0:
Hi! 👋

Version: `@convex-dev/agent@0.6.1`, `ai@6.0.142`

Thanks for all you do! It's been quite an experience moving to convex.

**My issue:**

When a tool call fails, AI SDK v6 returns it as a `tool-error` content part (not `tool-result`).

`willContinue` only counts step.toolResults, so it sees fewer results than calls and stops even though the step actually completed. This causes agents to stop mid-work when any tool errors occur during parallel execution.

in File: `src/client/utils.ts:13`

```
// line 13 — currently:
if (step.toolCalls.length > step.toolResults.length) return false;

// Ideally would be something like:
const completedOutputs = step.content?.filter(
  (p) => p.type === "tool-result" || p.type === "tool-error",
).length ?? step.toolResults.length;
if (step.toolCalls.length > completedOutputs) return false;
```

robelest:
Hi @impish0 — sorry for the delay, and thank you for reporting this and providing a clear explanation.

I re-verified this against `@convex-dev/agent@0.7.0` (`694721d`) and could not reproduce the continuation problem.

The local regression harness covered both sides of the behavior:

- a parallel step with one `tool-result` and one `tool-error` continued correctly
- a step with a genuinely missing tool output still waited

As a negative control, restoring the old `toolCalls.length > toolResults.length` check reproduced the original failure.

The relevant continuation logic was corrected as part of #304, which now counts both `tool-result` and `tool-error` content parts as completed outputs.

Could you confirm whether this still reproduces for you on the latest `0.7.x` release?

Issue #234, opened 2026-03-03
createUserUIMessage doesn't spread partCommon on string content path


Jake-qp:
## Bug

In `UIMessages.js`, `createUserUIMessage` doesn't spread `partCommon` (which carries `providerMetadata`) onto parts when the message has simple string content.

**Line 211** (string content path):
```js
parts: [{ type: "text", text: content }]  // partCommon NOT spread
```

**Line 216** (array content path):
```js
parts: content.map((part) => ({ ...partCommon, ...part }))  // partCommon IS spread ✓
```

Compare with `createSystemUIMessage` and `createAssistantUIMessage` which both correctly spread `partCommon` on all paths.

## Impact

When using `saveMessage` with `prompt` (string) + `metadata.providerMetadata`, the `providerMetadata` is stored in the DB but lost when constructing UIMessage parts. Consumers reading `parts[].providerMetadata` get `undefined`.

## Reproduction

```typescript
const { messageId } = await saveMessage(ctx, components.agent, {
  threadId,
  prompt: "Hello",  // string → hits line 211
  metadata: {
    providerMetadata: {
      myApp: { customField: "value" },
    },
  },
});

// Later, in useUIMessages:
// message.parts[0].providerMetadata → undefined (expected: { myApp: { customField: "value" } })
```

## Workaround

Use `message` with array content instead of `prompt`:

```typescript
const { messageId } = await saveMessage(ctx, components.agent, {
  threadId,
  message: {
    role: "user",
    content: [{ type: "text", text: "Hello" }],  // array → hits line 216
  },
  metadata: {
    providerMetadata: {
      myApp: { customField: "value" },
    },
  },
});
```

## Expected Fix

Spread `partCommon` on the string content path (line 211):

```js
parts: [{ ...partCommon, type: "text", text: content }]
```

## Version

`@convex-dev/agent@0.3.2`

robelest:
Hi @Jake-qp — sorry for the delay, and thank you for reporting this and providing a clear reproduction.

I confirmed the issue still exists in `@convex-dev/agent@0.7.0` and opened #318 with a fix. The PR applies the existing common text-part fields to string-backed user messages, so `providerMetadata` and `state` now match the array-content path.

Thanks again for contributing this report.

PR #260, opened 2026-05-09
Fix three UI / streaming / search bugs

sethconvex:
## Summary
Two small, independent bug fixes — each with the exact root cause already identified in the issue report.

- **#234** — `createUserUIMessage` was not spreading `partCommon` on the string-content path, so `providerMetadata` and the `streaming` / `done` state were dropped when a user message was saved with a plain string `prompt`. The array-content path already spread `partCommon`; the assistant and system variants of this function did so on both paths. Fix mirrors them.
- **#256** — `textSearch` applied the `targetMessage.order` upper-bound to *all* search hits, including messages from other threads when `searchAllMessagesForUserId` was set. Since `order` is only meaningful within a single thread, valid cross-thread matches were dropped whenever the new thread's current message had a lower `order`. The DB-level `lte(order)` predicate is now skipped for cross-thread searches, and the post-filter only enforces the order constraint within the same `threadId`.

Closes #257 

## Test plan
- [x] `npm test` — all 262 tests pass.
- [x] `src/toUIMessages.test.ts` updated: the user-message-with-string-content test was asserting the buggy shape (no `state`); now matches the assistant path which already expected `state: "done"`.
- [ ] Manual verification of #256 against a real Convex deployment with cross-thread search enabled (not done locally — repro requires populated cross-thread data).
- [ ] Manual verification of #234 with a `saveMessage({ prompt, metadata: { providerMetadata } })` call.

## Notes
- No public API changes.
- Originally also included a fix for #240 (`willContinue` ignoring `tool-error` parts); that fix landed in #241 and has been dropped from this branch via rebase.
- Pre-existing `npm run typecheck` errors in `example/convex/setup.test.ts` are present on `main` and unrelated to this change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,7 +40,7 @@
         "clsx": "2.1.1",
         "convex": "1.35.1",
         "convex-helpers": "0.1.114",
-        "convex-test": "0.0.44",
+        "convex-test": "0.0.51",
         "dayjs": "1.11.20",
         "dotenv": "16.6.1",
         "eslint": "9.39.4",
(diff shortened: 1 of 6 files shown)


sethconvex:
@codex review

ianmacartney (approved):
lgtm - I had started branch `issue-256` and will copy over one addition from there (where it does the same order fix, but for vector search too). I had also updated tests and fixed some lint errors will update this branch for you to inspect & land if you're happy with it

robelest:
All three are fixed on main, so closing.

- **#234**: `parts.push({ type: "text", text, ...partCommon })` at `src/vercel/UIMessages.ts:347`. That is the only production line in this PR.
- **#240**: `willContinue` counts both `tool-result` and `tool-error` at `src/vercel/client/utils.ts:35-38`, via #241.
- **#256**: fixed by #257 (6301f30), which also added 120 lines of coverage to `src/component/messages.test.ts`.

Coverage is in place for all three, so nothing is lost here.

PR #262, opened 2026-05-10
Fix typecheck failure in example/convex/setup.test.ts

sethconvex:
## Summary
Restores `npm run typecheck` to green. Main has been failing on every push since 2026-04-01, masking real type regressions in subsequent PRs.

## Root cause
Commit e29f2bb (\"better convex test types\") made `@convex-dev/agent`'s `register` signature generic. Sibling components didn't all follow:

| Package | Pinned here | Latest | Generic register? |
|---|---|---|---|
| `@convex-dev/agent` | this repo | — | ✅ |
| `@convex-dev/workflow` | 0.3.6 | **0.3.12** | ✅ in 0.3.12 |
| `@convex-dev/workpool` | 0.3.1 (transitive) | **0.4.6** | ✅ in 0.4.x |
| `@convex-dev/rate-limiter` | 0.3.2 | 0.3.2 (latest) | ❌ still non-generic |

`TestConvex` is invariant in its schema parameter (it appears in both contravariant input positions like `t.query(...)` args and covariant return positions), so a `TestConvex<SchemaDefinition<{ rawUsage: ..., invoices: ... }>>` is **not** assignable to `TestConvex<SchemaDefinition<GenericSchema, boolean>>` even though the value is structurally compatible.

## Fix
- Bump `@convex-dev/workflow` 0.3.6 → 0.3.12 (now generic).
- Add `@convex-dev/workpool` 0.4.6 as an explicit dev dep so the transitive workpool is upgraded; workflow's own `workpool.register(t, ...)` would otherwise fail typecheck against the older 0.3.1 workpool.
- Cast `t` only at the `rateLimiter.register(...)` call site — rate-limiter@0.3.2 is the latest published version and still has the non-generic signature. The cast is annotated as a temporary workaround; an upstream PR genericizes rate-limiter's register, and once a new release lands the cast can be deleted.

## Test plan
- [x] `npm run typecheck` — passes locally (was failing on main for ~6 weeks)
- [x] `npm test` — all 262 tests pass
- [x] `npm run lint` — clean

## Follow-up
- Upstream PR on `get-convex/rate-limiter` to genericize `register`. Once shipped, drop the cast.


ianmacartney (commented):
I think we can drop this if we update convex-test

PR #259, opened 2026-05-09
Add httpStreamText and useHttpStream

sethconvex:
## Summary

First-class HTTP streaming for `@convex-dev/agent`, plus a full demo:

- **`httpStreamText` / `httpStreamUIMessages`** (`src/client/http.ts`) — return a `Response` from a Convex `httpAction` backed by AI SDK's `toTextStreamResponse()` / `toUIMessageStreamResponse()`. Handles auth, thread auto-creation (with `userId`), optional delta persistence, and CORS headers. Sets `X-Message-Id` and `X-Stream-Id` response headers.
- **`Agent.asHttpAction()`** — convenience method that wraps `httpStreamText` with body parsing (`{ threadId?, prompt?, promptMessageId?, messages? }`), thread auto-creation, CORS, a `format: "text" | "ui-messages"` switch, and an optional `authorize` callback.
- **`useHttpStream`** (`src/react/useHttpStream.ts`) — React hook that POSTs to the endpoint, accumulates the streamed text, and exposes `streamId` / `messageId` so callers can pass them to `useUIMessages` via `skipStreamIds` for deduplication. Includes `abort()`.
- **`consumeTextStream` / `supportsStreaming`** (`src/react/httpStreamUtils.ts`) — small utilities the hook is built on; exported for users who want to roll their own.
- **`streamText` metadata** — `streamText`'s result now carries `streamId` when `saveStreamDeltas` is set; the streamId is created eagerly so the HTTP header is populated even when the response returns before consumption finishes.
- **Streaming demo** at `/streaming-demo` showing all three patterns (delta / HTTP / one-shot) side-by-side with abort, stream inspector, tool approval, and continuation. Wired up via `agent.asHttpAction()`.

This supersedes #232 — same shape, but rebased onto current `main` and incorporates the CodeRabbit feedback from that thread.

### CodeRabbit feedback from #232 incorporated

- `userId` is propagated to auto-created threads.
- `saveStreamDeltas: true` is normalized to `{ returnImmediately: true }` for HTTP flows so the response body actually streams.
- `consumeTextStream` listens for `signal.abort` and calls `reader.cancel()`, so a pending `read()` doesn't block the caller.
- `useHttpStream` tracks a `requestId` so a superseded request's `finally` can't flip `isStreaming=false` on a newer in-flight stream.

### Drive-by fix

[...]

Issue #181, opened 2025-10-30
Sub-messages flow problem


franckdsf:
When an **orchestrator agent** triggers a sub-agent (**photographer**) as a tool call, the sub-agent’s internal tool calls and message updates are not saved as part of the same mutation cycle.
Instead, they are created and then later **deleted and reinserted**, causing major flickering and message duplication in the UI.

---

### 💡 Expected Behavior

* The **orchestrator** triggers the **photographer** as a tool call.
* The **photographer** executes its sub-tools, and all messages (both from the orchestrator and photographer) are recorded as part of a **single message stream** (or not, but at least, not merged at the end).
* Once the **photographer** is done, the orchestrator finalizes the result cleanly, without message re-creation or flickering.

---

### 🚫 Actual Behavior

1. The **orchestrator** starts handling the request via `streamText`.
2. It triggers the **photographer** tool → message saved as `toolCall started`.
3. A **new message** is generated for the photographer’s tool calls.
4. Once the photographer is done:

   * The orchestrator’s status is set to `done`.
   * All photographer messages are **deleted**.
   * The same photographer messages are **re-saved** as part of the orchestrator’s final message content.

Because these steps occur sequentially, **each sub-step triggers its own render** instead of being part of a single mutation.
This produces heavy **UI flickering**, duplicated data, and message jumps.
Additionally, the orchestrator’s “done” status appears **before** messages are merged, causing inconsistent intermediate states.

---

#### Expected

```
1. Orchestrator handles request
2. Triggers photographer
3. Photographer triggers sub-tools (all messages saved under same message or own message)
4. Photographer returns result
5. Everything is tidy and clean
```

#### Actual

```
1. Orchestrator handles request
2. Triggers photographer (toolCall started)
3. Separate message chain created for photographer
4. Photographer completes
5. Orchestrator marked done
6. Photographer messages deleted & re-added → UI flickers
```

---

### Notes / Possible Fix

It seems the issue stems from **message synchronization** happening **after** the orchestrator finalizes.
Maybe the orchestrator’s `done` status and message merging should occur **within the same atomic mutation**, ensuring the whole process updates in one go.

Currently, I’m working around this by:

* Hooking message merging at runtime when the orchestrator isn’t done yet. (since it produces an orchestrator message AND a photographer message at the same time).
* Debouncing the orchestrator’s `done` status until merging finishes...

However, that’s only a partial fix.

PR #266, opened 2026-05-19
fix: persist final step when saveStreamDeltas.returnImmediately is true (#265)

### LINKED ISSUE #265: saveStreamDeltas with returnImmediately: true never saves final step to messages table
## Bug

When using `agent.streamText()` with `saveStreamDeltas: { returnImmediately: true }`, the final step's message is never persisted to the messages table. Intermediate steps (tool calls) are saved correctly, but the final text response is lost.

## Root cause (in `streamText.js`)

1. `onStepFinish` for the final step defers saving: it sets `pendingFinalStep = step` and calls `streamer.markFinishedExternally()` (lines 79-80)
2. The code that executes the deferred save (lines 114-118: `if (pendingFinalStep && streamer) { ... }`) runs **before** `onStepFinish` fires — because `returnImmediately: true` skips `await stream` at lines 89-113
3. `consumeStream` finishes later in the background but skips calling `finish()` because `#finishedExternally` is `true`
4. **Result:** `pendingFinalStep` is never set when lines 114-118 execute, so `call.save` for the final step never runs. The stream is never marked finished either.

## Reproduction

\`\`\`ts
const result = await agent.streamText(
  ctx,
  { threadId },
  { promptMessageId },
  {
    saveStreamDeltas: {
      returnImmediately: true,
      chunking: 'word',
      throttleMs: 100,
    },
  },
);

const response = result.toTextStreamResponse();
// Client receives full streaming text via HTTP...
// But listUIMessages() afterwards is missing the assistant's text response
\`\`\`

## Expected

All steps (including final text response) are saved to the messages table, and the stream is marked as finished.

## Actual

Only intermediate steps are saved. The final text response is never written to the messages table. The stream stays in streaming status indefinitely.

## Workaround

Remove `saveStreamDeltas` entirely. Without it, `onStepFinish` always takes the `else` branch and calls `call.save()` for every step including the final one. HTTP streaming via `toTextStreamResponse()` still works.

## Suggested fix

When `returnImmediately: true`, the deferred save at lines 114-118 needs to await `onStepFinish` completion, or the final step should not be deferred (save it immediately in `onStepFinish` like intermediate steps).

sethconvex:
## Summary

Fixes #265: when `streamText` is called with `saveStreamDeltas: { returnImmediately: true }`, the final assistant message was never persisted and the stream stayed stuck in `"streaming"` status forever. Only intermediate (tool-call) steps were saved.

## Root cause

The deferred-save path added for atomic stream finish (#181) doesn't work when `streamText` returns without awaiting stream consumption:

1. `onStepFinish` runs only when the caller later drains the stream (e.g. via `result.toTextStreamResponse()`).
2. With `returnImmediately: true`, `streamText` skips its own `await stream` block and exits.
3. The post-await `if (pendingFinalStep && streamer)` block runs synchronously after exit — but `onStepFinish` hasn't fired yet, so `pendingFinalStep` is still `undefined`.
4. The caller drains the stream later, `onStepFinish` fires, sets `pendingFinalStep` and calls `markFinishedExternally()` — but no one is reading `pendingFinalStep` anymore, and `consumeStream` skips `finish()` due to the external-finish flag.
5. Net effect: final message dropped, stream stuck.

## Fix

Branch on a new `willAwaitStream` flag in `onStepFinish`:

- If `streamText` will await consumption (`saveStreamDeltas === true` or `returnImmediately: false`): defer the save via `pendingFinalStep` as before (preserves the atomic-finish behavior from #181).
- If not (`returnImmediately: true`): save the final step **inline** using `streamer.getOrCreateStreamId()` so the message persistence and stream-finish land atomically — before `streamText` returns.

Also guards `DeltaStreamer.addParts` against late deltas after `markFinishedExternally`. The stream record is already in `"finished"` state and `streams.addDelta` would silently drop those writes anyway; skipping them is just an optimization, but it avoids racing the inline save.

## Provenance

These two pieces (the `willAwaitStream` branch and the `addParts` guard) are lifted from the open PR #259 ("Add httpStreamText and useHttpStream"), which bundles them with a larger HTTP streaming surface (~2,800 LOC across 21 files). This PR cherry-picks just the bug fix so #265 can ship independently of that feature work — which is also currently blocked on merge conflicts with `main`.

## Notes for reviewers

- **Minor behavioral nuance worth knowing:** with the new `addParts` guard, delta chunks emitted by the model after `onStepFinish` fires but before `consumeStream` has finished pumping them are silently dropped on the `returnImmediately` path. The persisted **message** is unaffected — it carries the complete final-step text via `call.save({step})`. A UI subscribed to live deltas could see slightly-truncated streaming content until it reconciles against the saved message, but no canonical data is lost.
- The fix preserves the #181 atomic-finish guarantee for the original path (when we do await stream consumption).
- The full HTTP streaming PR #259 should still land for its other improvements, but is intentionally out of scope here.

## Test plan

- [x] New regression test `src/client/streamText.test.ts` reproduces the issue: calls `streamText` with `returnImmediately: true`, drains the result, then asserts (a) a persisted assistant message exists with the model's text and (b) no stream is left in `"streaming"` status.
- [x] Test fails on `main` with `expected 0 to be greater than 0` (no assistant message persisted).
- [x] Test passes with the fix applied.
- [x] `npm run lint` clean.
- [x] `npx tsc --noEmit` clean.
- [x] Full vitest suite: 257/257 unit tests pass. (The 3 `example/convex/*.test.ts` failures are pre-existing on `main` — they need `npm run build` to populate `dist/` first; unrelated to this change.)

### DIFF
--- a/example/convex/setup.test.ts
+++ b/example/convex/setup.test.ts
@@ -1,6 +1,7 @@
 /// <reference types="vite/client" />
 import { test } from "vitest";
-import { convexTest } from "convex-test";
+import { convexTest, type TestConvex } from "convex-test";
+import type { GenericSchema, SchemaDefinition } from "convex/server";
 import schema from "./schema.js";
 import agent from "@convex-dev/agent/test";
 import workflow from "@convex-dev/workflow/test";
@@ -10,8 +11,15 @@ export const modules = import.meta.glob("./**/*.*s");
 export function initConvexTest() {
   const t = convexTest(schema, modules);
   agent.register(t);
-  workflow.register(t);
-  rateLimiter.register(t);
+  // workflow and rate-limiter still type their `register` against the
+  // pre-generic SchemaDefinition<GenericSchema, boolean>. Cast through
+  // the broader type so this file typechecks regardless of which version
+  // of those packages is resolved.
+  const generic = t as unknown as TestConvex<
+    SchemaDefinition<GenericSchema, boolean>
+  >;
+  workflow.register(generic);
+  rateLimiter.register(generic);
   return t;
 }
 
--- a/src/client/streamText.test.ts
+++ b/src/client/streamText.test.ts
@@ -1,0 +1,102 @@
+import { describe, expect, test } from "vitest";
+import { Agent, createThread } from "./index.js";
+import {
+  defineSchema,
+  type DataModelFromSchemaDefinition,
+  type ApiFromModules,
+  type ActionBuilder,
+  actionGeneric,
+  anyApi,
+} from "convex/server";
+import { v } from "convex/values";
+import { components, initConvexTest } from "./setup.test.js";
+import { mockModel } from "./mockModel.js";
+
+const schema = defineSchema({});
+type DataModel = DataModelFromSchemaDefinition<typeof schema>;
+const action = actionGeneric as ActionBuilder<DataModel, "public">;
+
+const FINAL_TEXT = "Hello from the model";
+
+const agent = new Agent(components.agent, {
+  name: "stream-test",
+  languageModel: mockModel({
+    content: [{ type: "text", text: FINAL_TEXT }],
+  }),
+});
+
+// Action that exercises streamText with saveStreamDeltas.returnImmediately=true.
+// It consumes the stream after streamText returns, simulating the HTTP response
+// path described in issue #265.
+export const streamTextReturnImmediately = action({
+  args: { threadId: v.string() },
+  handler: async (ctx, { threadId }) => {
+    const result = await agent.streamText(
+      ctx,
+      { threadId },
+      { prompt: "Test" },
+      {
+        saveStreamDeltas: {
+          returnImmediately: true,
+          chunking: "word",
+          throttleMs: 0,
+        },
+      },
+    );
+    // Drain the stream the way an HTTP response would. This triggers
+    // onStepFinish for every step, including the final one.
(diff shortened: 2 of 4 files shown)


ianmacartney (approved):

sethconvex:
Pushed `b602a46` addressing two things:

**CodeRabbit's `#sendDelta` race (valid, fixed)** — confirmed in code at `src/client/streaming.ts:351`. An in-flight `#sendDelta` started before `markFinishedExternally()` would call `addDelta`, get `success === false` because the row is already `finished`, and the current path would `onAsyncAbort + abort()` — turning a successfully externally-finished stream into the failure path. Fixed by skipping the abort branch when `#finishedExternally` is set; the late-write miss is benign.

**CI failure (unrelated to this PR, but fixed here)** — the `Test and lint` failure on commit `10cd3cf` was in the `tsc -p example/convex` step, complaining about `@convex-dev/workflow/test` and `@convex-dev/rate-limiter/test` typing `register` against the pre-generic `SchemaDefinition<GenericSchema, boolean>`. Last 3 CI runs on `main` itself failed with the same error (runs `25619373541`, `24437925179`, `24047631068`) — it's a pre-existing break, not introduced here. Included the drive-by typecheck cast from PR #259 to get this PR's CI green.

Local verification: `npm run lint` clean, `npm run typecheck` clean, `npx vitest run` 257/257 passing (streaming tests included).

PR #277, opened 2026-06-02
use ActionCtx versions of runQuery/runMutation for compatibility

ianmacartney:
In the future the `ctx.runQuery` of a QueryCtx / MutationCtx may differ from ActionCtx by adding extra options (e.g. snapshot query, transaction limits). For now, the ActonCtx is the more generic version, so if the function accepts that, the user can pass any of QueryCtx/MutationCtx/ActionCtx and it should work.

This is only really an issue for Workflow, where it presents as a ctx that has ActionCtx-like runQuery/runMutation/runAction, but does not have auth/storage, so doesn't match the ActionCtx here.

This change allows workflow's step (ctx) to be passed to agent functions. A follow-up (#279) could explicitly handle WorkflowCtx and pass { inline: true } for various places where it makes sense, to explicitly optimize for workflows


ianmacartney:
* **#279** <a href="https://app.graphite.com/github/pr/get-convex/agent/279?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#277** <a href="https://app.graphite.com/github/pr/get-convex/agent/277?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/agent/277?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #279, opened 2026-06-02
use inline: true in workflows

ianmacartney:
For lightweight operations, run queries / mutations inline in Workflows, to avoid the workpool overhead.

This enables using Workflow + Agent more idiomatically, creating threads & saving messages before kicking out to an action to do generation, which likely saves ~1s or more per workflow

Risk:

1. adds a peer dependency on Workflow
2. If they had workflowId on ctx incidentally and had some custom ctx.run* functions that didn't expect an extra param (unlikely)
3. Maybe they expected it to run in a separate transaction and that was impotant for some reason?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -75,6 +75,7 @@
       },
       "peerDependencies": {
         "@ai-sdk/provider-utils": "^4.0.6",
+        "@convex-dev/workflow": "^0.4.3",
         "ai": "^6.0.35",
         "convex": "^1.24.8",
         "convex-helpers": "^0.1.103",
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
   },
   "peerDependencies": {
     "@ai-sdk/provider-utils": "^4.0.6",
+    "@convex-dev/workflow": "^0.4.3",
     "ai": "^6.0.35",
     "convex": "^1.24.8",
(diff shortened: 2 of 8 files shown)


ianmacartney:
* **#279** <a href="https://app.graphite.com/github/pr/get-convex/agent/279?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/agent/279?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#277** <a href="https://app.graphite.com/github/pr/get-convex/agent/277?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

robelest:
small consistency q: should the other lightweight helpers that accept
`MutationCtx` use the new wrappers too?

Right now `createThread` / `saveMessages` go inline in workflows, but
`listMessages`, `searchThreadTitles`, and delete helpers still call `ctx.run*`
directly. Maybe intentional, but it feels a little mixed.

ianmacartney:
search might require vector search, which isn't available in workflows. listing messages I guess is harmless - it's not reading more data than it's returning, so may as well inline it

robelest (approved):
LGTM

robelest:
@cubic review

PR #287, opened 2026-06-24
Add opt-in Anthropic compaction for long threads

hpaul:
Hey! 👋

Long threads on Anthropic models eventually run past the context window, and right now the only real lever is `recentMessages` truncation — which just throws old context away instead of summarizing it.

Turns out Anthropic's server-side compaction (`compact_20260112`) plugs into the agent's existing storage really cleanly, so this ended up being a small change! The summary comes back as a text part tagged `providerOptions.anthropic.type === "compaction"`, and `serializeMessage`/`toModelMessage` already round-trip that untouched — nothing to change in the mapping/validators layer.

### What's in here
- Bumps `@ai-sdk/anthropic` 3.0.13 → 3.0.62 (compaction landed in 3.0.41).
- Adds an opt-in `contextOptions.compaction` knob:
  ```ts
  contextOptions: { compaction: { triggerTokens: 50_000 } } // + optional instructions
  ```
  When set it merges the compact edit into `providerOptions.anthropic` (preserving anything you already pass, and it won't double-add), and widens the recent-message window so history actually reaches the trigger and the stored summary gets replayed instead of re-summarized.
- Fully backward compatible — leave it unset and behavior is exactly the same as today.

### How I checked it (live, not hand-wavy)
`npm run verify:compaction` drives a ~115k-token thread past the trigger, round-trips the response through the real `serializeMessage`/`toModelMessage`, resends, and reads `usage.iterations`: turn 2 shows **no** re-compaction (the message pass stays tiny), proving the round-tripped summary was honored by the API. Also added unit tests for the merge logic.

Totally happy to adjust naming or placement (e.g. whether `compaction` belongs on `contextOptions` or its own option) — let me know what fits the codebase best! 🙏


hpaul:
@ianmacartney @robelest Do you think this could be merged? Should I add something more?

ianmacartney:
@robelest what do you think about this now that we're in v7 territory?

PR #295, opened 2026-07-15
Streaming chat ergonomics: typed streaming-query helpers, validators, and known-good AI SDK versions

sethconvex:
## Why

An AI coding-eval for the Convex plugin asked an agent to build a streaming chat app on `@convex-dev/agent`. It took 69 turns and **never compiled**, hitting two concrete walls that any from-scratch user can hit:

### 1. The streaming return type is a landmine (TS2719)

The docs and the `useUIMessages` hook steer you to `listUIMessages` (which returns a page of `UIMessage`s), but the only exported `returns` validator, `vStreamMessagesReturnValue`, has a `vMessageDoc`-shaped page. Combining them, as convex best practices ("always add return validators") encourage, produces:

```
error TS2719: Type '(ctx: ...) => ...' is not assignable to type '(ctx: ...) => ...'. Two different types with this name exist, but they are unrelated.
  Type 'Promise<{ streams: SyncStreamsReturnValue; page: UIMessage[]; ... splitCursor?: string | null | undefined; pageStatus?: "SplitRecommended" | ... }>' is not assignable to type 'ValidatorTypeToReturnType<{ splitCursor?: ...; pageStatus?: "SplitRecommended" | "SplitRequired" | null | undefined; streams?: ...; page: ...MessageDoc... }>'
```

The eval agent burned ~20 turns grep/sed-ing `.d.ts` files trying to reconstruct the exact split-pagination validator shape. There was no exported UIMessage validator to reach for.

A second, hidden trap compounded it: even a *correct* streaming query with a `returns` validator was rejected by the hooks, because `v.optional(streams)` infers an optional property (`streams?:`) while `StreamQuery` (the type behind `useUIMessages(..., { stream: true })`) required a non-optional `streams`, yielding:

```
error TS2322: Type 'true' is not assignable to type 'ErrorMessage<"To enable streaming, your query must take in streamArgs: vStreamArgs and return a streams object returned from syncStreams. See docs.">'
```

### 2. Peer-dependency guessing

The component requires `ai@^6` + `@ai-sdk/provider-utils@^4`, but nothing documented which **provider** majors pair with it. The eval agent installed `@ai-sdk/anthropic@4` (wrong AI SDK generation), got an unresolvable peer conflict, and thrashed reinstalling before stumbling onto `@ai-sdk/anthropic@^3`.

## What changed (all additive)

**New server helpers** (`src/client/streaming.ts`, exported from the package root):

- `listUIMessagesWithStreams(ctx, component, { threadId, paginationOpts, streamArgs })` - fetches the paginated UIMessages *and* the stream deltas, returning exactly what `useUIMessages(..., { stream: true })` expects.
- `listMessagesWithStreams(...)` - same for `MessageDoc`s / `useThreadMessages`.

**New validators**:

- `vUIMessage` (`src/UIMessages.ts`) - validator matching the `UIMessage` type (AI SDK `parts`/`metadata` are validated as `any` since they're generic over the app's tools).
- `vStreamUIMessagesReturnValue` - the `returns` validator matching `listUIMessagesWithStreams`.
- `vSyncStreamsReturnValue` - the `streams` field alone, for hand-rolled queries. (`vStreamMessagesReturnValue` is unchanged in shape, now built from it.)

**One targeted fix** (`src/react/types.ts`): `StreamQuery`'s return type now has `streams?:` optional, so queries declaring a `returns` validator satisfy the `stream: true` guard. Queries that don't return `streams` at all still fail it (no properties in common), and a type test pins that.

**Docs**: README gains a complete "Streaming chat quickstart" (server query + mutation + action + React hook that all type-check), and README + `docs/getting-started.mdx` document the version pairing: **`ai@^6` pairs with `@ai-sdk/*` providers at `^3`** (e.g. `@ai-sdk/anthropic@^3`), matching this repo's own pinned devDeps (`ai@6.0.35` + `@ai-sdk/anthropic@3.0.13`). `docs/streaming.mdx` and the `useUIMessages` docstring now lead with the helper.

**Example**: `example/convex/chat/streaming.ts`'s `listThreadMessages` now uses the helper + `returns` validator (so the example exercises the exact new path end-to-end, including `optimisticallySendMessage`); the manual composition is preserved as `listThreadMessagesCustom`.

## Before / after for a from-scratch user

Before (what the eval agent kept trying, TS2719):

```ts
export const listThreadMessages = query({
  args: { threadId: v.string(), paginationOpts: paginationOptsValidator, streamArgs: vStreamArgs },
  returns: vStreamMessagesReturnValue, // MessageDoc-shaped: mismatch
  handler: async (ctx, args) => {
    const paginated = await listUIMessages(ctx, components.agent, args); // UIMessage-shaped
    const streams = await syncStreams(ctx, components.agent, args);
    return { ...paginated, streams }; // TS2719
  },
});
```

After:

```ts
export const listThreadMessages = query({
  args: { threadId: v.string(), paginationOpts: paginationOptsValidator, streamArgs: vStreamArgs },
  returns: vStreamUIMessagesReturnValue,
  handler: async (ctx, args) => {
    await authorizeThreadAccess(ctx, args.threadId);
    return listUIMessagesWithStreams(ctx, components.agent, args);
  },
});
```

```tsx
const { results: messages } = useUIMessages(
  api.chat.listThreadMessages,
  { threadId },
  { initialNumItems: 10, stream: true },
);
```

## Verification

- `npm run build` - clean
- `npm run typecheck` (`tsc --noEmit && tsc -p example && tsc -p example/convex`) - clean
- `npm run test` (`vitest run --typecheck`) - **25 files, 275 tests passed, no type errors** (was 24 files / 272 tests before; adds `src/client/listWithStreams.test.ts` with runtime `parse()` validation of both helpers against both validators for `kind: "list"`, `kind: "deltas"`, and no-`streamArgs` calls, plus type-level assertions that the helper return types satisfy the validators, that both validator-typed and inferred-typed queries extend `StreamQuery`/`UIMessagesQuery`, and that a non-streaming query still does **not** extend `StreamQuery`)
- `npm run lint` - clean
- Reproduced the original TS2719 verbatim on this codebase before the fix, and confirmed the same query now compiles.

Surfaced by a Convex plugin coding-eval session.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016XVeiChqZyhVRXu34va31k


robelest:
Merged `main` in, so this is mergeable again.

Two changes on top of your work:

- **Version docs updated** to `ai@^7`, providers at `^4`, `provider-utils@^5`. They said v6 with providers at `^3`, which was right in July.
- **`listWithStreams.test.ts` moved** to `src/vercel/client/`, with two import paths fixed. Git tracked the other renames, but that file was new here so it had nothing to rename from.

PR #298, opened 2026-07-22
feat(vector): support 2560-dimensional embeddings

JustinMissmahl:
## Summary

Add `2560` to the vector dimensions supported by the Agent component.

This enables embedding models that return 2560-dimensional vectors, including:

- `perplexity/pplx-embed-v1-4b`
- `perplexity/pplx-embed-context-v1-4b`

## Motivation

The Agent component currently restricts vector dimensions to a predefined
allowlist. Although Convex vector indexes support 2560 dimensions, it is
missing from that list.

As a result, applications cannot use these embedding models at their native
dimension without maintaining a local patch or reducing the embedding size.

## Changes

- Add `2560` to `VectorDimensions`
- Regenerate the Convex component types to include the new dimension

## Testing

- Ran `npm run build:clean`
- Ran `npm run test`
- Ran `npm run lint`
- Ran `npm run typecheck`

I have also been using this change successfully through a local package patch
with `perplexity/pplx-embed-v1-4b`.

---

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

### DIFF
--- a/src/component/_generated/component.ts
+++ b/src/component/_generated/component.ts
@@ -148,6 +148,7 @@ export type ComponentApi<Name extends string | undefined = string | undefined> =
               | 1408
               | 1536
               | 2048
+              | 2560
               | 3072
               | 4096;
             model: string;
@@ -4770,6 +4771,7 @@ export type ComponentApi<Name extends string | undefined = string | undefined> =
               | string
               | string
               | string
+              | string
             >;
           },
           null,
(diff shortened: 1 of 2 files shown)


ianmacartney (approved):
thanks! Sounds like it's worth adding

ianmacartney:
I'll defer to @robelest whether we should wait to land it or replay the change on top of some big structural PRs he's working with, but `npm i https://pkg.pr.new/@convex-dev/agent@298` should allow you to try it out (sounds like you have a patched version already though)

PR #301, opened 2026-07-30
feat(vercel)!: upgrade AI SDK integration to v7

robelest:
Stacked on upstream #302 (`robel/agent-core-boundary`). Review the eight commits above that base.

Upgrades the Vercel adapter to AI SDK 7 (`ai@7`, provider utilities v5, and official provider packages v4) while keeping SDK-specific behavior behind the `src/vercel/` boundary.

## What changed

- Migrates prompts, callbacks, usage, tools, runtime context, raw-response opt-ins, provider models, and string registry model IDs to AI SDK 7.
- Writes new streams as `UIMessageChunkV7` while retaining the pinned AI SDK 6 decoder for aborted or in-flight `UIMessageChunk` rows.
- Persists and round-trips AI SDK 7 `custom` and `reasoning-file` assistant parts through model messages, UI messages, and server-side stream recovery.
- Supports canonical tool-result `{ type: "file" }` content with tagged data, URL, provider-reference, and inline-text variants.
- Routes oversized binary reasoning/tool-result files through the existing tracked Convex storage and reference-count lifecycle.
- Fixes transient data chunks incorrectly anchoring orphan continuation streams.
- Updates the migration guide and tool examples, and makes the release command produce `0.7.0` from the current `0.6.x` line.

## Persisted-format compatibility

No database migration is required. `UIMessageChunk` remains pinned to the AI SDK 6 wire semantics that originally wrote those rows; `UIMessageChunkV7` uses AI SDK 7 semantics. The format marker is always selected explicitly rather than inferred from chunk shape.

The canonical message schema changes are additive. The generated component binding diff is larger because Convex expands the updated message union into every public function that accepts or returns messages.

## External preview dependency

`@convex-dev/rag` remains pinned to the immutable preview build at `2bd0281`; the latest stable RAG release still targets AI SDK 6.

## Verification

- 323/323 tests pass with Vitest type checking
- root, example, and example/convex TypeScript projects pass
- ESLint passes
- Convex component codegen and package build pass


robelest:
Superseded by the clean native stack #304 → #305 → #306 → #307. The former combined v7 diff is now split into the adapter migration (#305), streaming format (#306), and durable persistence (#307).

PR #306, opened 2026-07-30
feat(vercel)!: migrate runtime writer to AI SDK 7

robelest:
## Summary

- Upgrade the Vercel adapter and examples to AI SDK 7 across prompts, callbacks, usage, tools, approvals, runtime context, provider models, and registry model IDs.
- Write AI SDK 7 chunks through the single persisted `UIMessageChunk` reducer/projector introduced by #305.
- Preserve existing saved message and stream data without retaining an AI SDK 6 runtime.
- Make stream cancellation terminal and race-safe for pre-aborted signals, in-flight stream creation, failed delta writes, and overlapping cleanup.
- Add the AI SDK 7 migration guide and 0.7.0 release notes.

## Release boundary

With #304–#306 merged, Agent is ready for the AI SDK 7 runtime release. A dedicated release commit will bump the package and lockfile from 0.6.4 to 0.7.0.

This release does not yet offload every oversized inline file encoding. Reasoning files, nested tool-result files, and tagged/base64/data-URL payloads will gain durable persistence in #307.

## Dependency status

This PR pins `@convex-dev/rag@0.8.0-alpha.0`, whose published peer dependency is `ai@^7.0.0`. The lockfile contains one AI SDK 7 dependency graph and no nested AI SDK 6 runtime.

## Validation

- Typecheck
- Lint
- Build
- 29 test files / 327 tests
- Two independent final reviews: clean

## Stack

1. #304 — isolate the provider adapter behind a neutral stream codec
2. #305 — extend the codec for the AI SDK 7 chunk superset
3. **#306 — migrate the runtime to AI SDK 7**
4. #307 — add durable persistence for oversized inline files

### DIFF
--- a/src/validators.ts
+++ b/src/validators.ts
@@ -15,6 +15,11 @@ export const vProviderMetadata = vProviderOptions;
 const providerMetadata = providerOptions;
 export type ProviderMetadata = Infer<typeof providerMetadata>;
 
+export const vProviderReference = v.object({
+  type: v.literal("reference"),
+  reference: v.record(v.string(), v.string()),
+});
+
 export const vThreadStatus = v.union(
   v.literal("active"),
   v.literal("archived"), // unused
@@ -42,7 +47,7 @@ export const vTextPart = v.object({
 
 export const vImagePart = v.object({
   type: v.literal("image"),
-  image: v.union(v.string(), v.bytes()),
+  image: v.union(v.string(), v.bytes(), vProviderReference),
   mediaType: v.optional(v.string()),
   /** @deprecated Use `mediaType` instead. */
   mimeType: v.optional(v.string()),
@@ -51,7 +56,7 @@ export const vImagePart = v.object({
 
 export const vFilePart = v.object({
   type: v.literal("file"),
-  data: v.union(v.string(), v.bytes()),
+  data: v.union(v.string(), v.bytes(), vProviderReference),
   filename: v.optional(v.string()),
   mediaType: v.optional(v.string()),
   /** @deprecated Use `mediaType` instead. */
@@ -88,7 +93,6 @@ export const vCustomContentPart = v.object({
   providerOptions,
   providerMetadata,
 });
-
 export const vRedactedReasoningPart = v.object({
   type: v.literal("redacted-reasoning"),
   data: v.string(),
@@ -198,6 +202,21 @@ export const vToolResultOutput = v.union(
           text: v.string(),
           providerOptions,
         }),
+        v.object({
+          type: v.literal("file"),
+          data: v.union(
+            v.object({ type: v.literal("url"), url: v.string() }),
+            vProviderReference,
+            v.object({
+              type: v.literal("data"),
+              data: v.union(v.string(), v.bytes()),
+            }),
+            v.object({ type: v.literal("text"), text: v.string() }),
+          ),
+          mediaType: v.string(),
+          filename: v.optional(v.string()),
+          providerOptions,
+        }),
         /** @deprecated Use `image-data` or `file-data` instead. */
         v.object({
           type: v.literal("media"),
@@ -219,6 +238,7 @@ export const vToolResultOutput = v.union(
         v.object({
           type: v.literal("file-url"),
           url: v.string(),
+          mediaType: v.optional(v.string()),
           providerOptions,
         }),
         v.object({
@@ -235,6 +255,11 @@ export const vToolResultOutput = v.union(
           providerOptions,
         }),
         v.object({
+          type: v.literal("file-reference"),
+          providerReference: v.record(v.string(), v.string()),
+          providerOptions,
+        }),
+        v.object({
           type: v.literal("image-data"),
           data: v.string(),
           /**
@@ -242,6 +267,11 @@ export const vToolResultOutput = v.union(
            * @see https://www.iana.org/assignments/media-types/media-types.xhtml
            */
           mediaType: v.string(),
+          providerOptions,
+        }),
+        v.object({
+          type: v.literal("image-file-reference"),
+          providerReference: v.record(v.string(), v.string()),
           providerOptions,
         }),
         v.object({
@@ -457,15 +487,34 @@ export const vFinishReason = v.union(
 );
 
 export const vUsage = v.object({
-  promptTokens: v.number(),
-  completionTokens: v.number(),
-  totalTokens: v.number(),
+  promptTokens: v.optional(v.number()),
+  completionTokens: v.optional(v.number()),
+  totalTokens: v.optional(v.number()),
   reasoningTokens: v.optional(v.number()),
   cachedInputTokens: v.optional(v.number()),
+  nonCachedInputTokens: v.optional(v.number()),
+  cacheWriteInputTokens: v.optional(v.number()),
+  textOutputTokens: v.optional(v.number()),
+  raw: v.optional(v.record(v.string(), v.any())),
 });
 export type Usage = Infer<typeof vUsage>;
 
 export const vLanguageModelCallWarning = v.union(
+  v.object({
+    type: v.literal("unsupported"),
+    feature: v.string(),
+    details: v.optional(v.string()),
+  }),
+  v.object({
+    type: v.literal("compatibility"),
+    feature: v.string(),
+    details: v.optional(v.string()),
+  }),
+  v.object({
+    type: v.literal("deprecated"),
+    setting: v.string(),
+    message: v.string(),
+  }),
   v.object({
     type: v.literal("unsupported-setting"),
     setting: v.string(),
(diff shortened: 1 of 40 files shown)


src/client/approval.test.ts
```diff
@@ -206,6 +226,28 @@ export const testDenyFlow = action({
...
+export const testApproveStreamFlow = action({
+  args: {},
+  handler: async (ctx) => {
+    const { thread } = await approvalStreamAgent.createThread(ctx, {
+      userId: "u-stream-approval",
+    });
+    const first = await thread.generateText({ prompt: "Delete stream.txt" });
+    const approvalId = getApprovalIdFromSavedMessages(first.savedMessages);
+    const { messageId } = await ctx.runMutation(
+      anyApi["approval.test"].submitApprovalForApprovalStreamAgent,
```
ianmacartney: Can we avoid using `anyApi` here?

src/client/approval.test.ts
```diff
@@ -206,6 +226,28 @@ export const testDenyFlow = action({
   },
 });
 
+export const testApproveStreamFlow = action({
+  args: {},
+  handler: async (ctx) => {
```
ianmacartney: we likely need to annotate the return value to avoid type cycles

src/client/approval.test.ts
```diff
@@ -393,6 +435,13 @@ export const submitApprovalForApprovalAgent = mutation({
   },
 });
 
+export const submitApprovalForApprovalStreamAgent = mutation({
+  args: { threadId: v.string(), approvalId: v.string() },
+  handler: async (ctx, args) => {
+    return approvalStreamAgent.approveToolCall(ctx, args);
```
ianmacartney: could we have inlined this into the action? I suspect we don't need this to have called a mutation first

src/vercel/client/definePlaygroundAPI.ts
```diff
@@ -282,19 +283,17 @@ export function definePlaygroundAPI<DataModel extends GenericDataModel>(
...
           component,
           step,
           {
             model: getModelName(agent.options.languageModel),
             provider: getProviderName(agent.options.languageModel),
           },
-          previousResponseMessageCount,
+          step.response.messages,
         );
-        previousResponseMessageCount = step.response.messages.length;
```
ianmacartney, line 297: so this bug was fixed? or we just fixed it in `serializeResponseMessages`

src/vercel/fileMaterialization.ts
```diff
@@ -0,0 +1,158 @@
+/**
+ * Storage effects for the Vercel adapter's otherwise storage-neutral codec.
+ *
+ * This module is intentionally one-way: it depends on the pure mapping
+ * planner, while mapping never imports or re-exports this module.
+ */
```
ianmacartney: I don't love the naming here - what is the intent behind it?

src/vercel/client/start.ts
```diff
@@ -185,34 +219,55 @@ export async function startGeneration<
...
+    ).experimental_include ?? {}),
+    ...((argsWithoutSystem as { include?: Record<string, boolean> }).include ??
+      {}),
+  };
+  const include = opts.rawRequestResponseHandler
+    ? {
+        ...requestedInclude,
+        requestBody: requestedInclude.requestBody ?? true,
+        ...(_operation === "generateText" || _operation === "generateObject"
+          ? { responseBody: requestedInclude.responseBody ?? true }
```
ianmacartney: what is this about?

src/vercel/client/streaming.ts
```diff
@@ -232,7 +232,11 @@ export class DeltaStreamer<T> {
       model?: string;
       provider?: string;
       providerOptions?: ProviderOptions;
-      format: "UIMessageChunk" | "TextStreamPart" | undefined;
+      format:
+        | "UIMessageChunk"
+        | "UIMessageChunkV7"
```
ianmacartney: reminder: if we can fold this into UIMessageChunk, we're remove it here

src/vercel/client/streamText.ts
```diff
@@ -136,7 +165,28 @@ export async function streamText<
...
+        await call.fail(reason);
+      } catch (error) {
+        internalError = error;
+      }
+      try {
+        await streamer?.fail(reason);
+      } catch (error) {
+        internalError ??= error;
+      }
+      await streamTextArgs.onAbort?.(event);
```
ianmacartney: can you double-check if we need to call all three of these separately or if there's any coupling where we should only call one or two of them?

src/vercel/client/streamText.ts
```diff
@@ -150,29 +200,46 @@ export async function streamText<
       }
       return undefined;
     },
-    onStepFinish: async (step) => {
+    onStepEnd: async (step) => {
       steps.push(step);
-      const createPendingMessage = await willContinue(steps, args.stopWhen);
+      const createPendingMessage = await willContinue(
+        steps as any,
+        args.stopWhen as any,
```
ianmacartney: this is fishy - why did types break?

src/vercel/react/useStreamingUIMessages.ts
```diff
@@ -139,18 +144,41 @@ export function useStreamingUIMessages<
...
+                    base as UIMessage,
+                    newParts,
+                    existing?.v7StreamState,
+                  )
+                : applyUIMessageChunksIncremental(
+                    base as UIMessage,
+                    newParts,
+                    prevState,
+                  );
+            const { message } = applied;
```
ianmacartney: reminder: follow up on V6 support for chunks

src/vercel/mapping.ts
```diff
@@ -72,25 +75,35 @@ export type SerializedContent = Message["content"];
...
+  filename?: string;
+  replaceWithUrl: (url: string) => void;
+};
+
+export type PlannedMessage = {
+  message: SerializedMessage;
+  files: PlannedFile[];
+};
+
+export function planMessage(
```
ianmacartney: Can you briefly explain what "planned" things are?

src/vercel/mapping.ts
```diff
@@ -270,102 +249,61 @@ export function serializeWarnings(
...
-  ctx: ActionCtx,
-  component: AgentComponent,
+export function planResponseMessages<TOOLS extends ToolSet>(
   step: StepResult<TOOLS>,
   model: ModelOrMetadata | undefined,
-  previousResponseMessageCount: number,
-): Promise<{ messages: MessageWithMetadata[] }> {
-  const newMessages = step.response.messages.slice(
-    previousResponseMessageCount,
-  );
```
ianmacartney, line 336: did this logic translate anywhere? I don't see it. Or is V7 better about this?

src/mapping.test.ts
```diff
@@ -3,14 +3,16 @@ import {
...
   toModelMessageDataOrUrl,
-  serializeMessage,
-  serializeNewMessagesInStep,
-  serializeResponseMessages,
   toModelMessage,
-  serializeContent,
+  serializeUsage,
   toModelMessageContent,
   autoDenyUnresolvedApprovals,
 } from "./vercel/mapping.js";
```
ianmacartney: is this test file supposed to be in ./vercel?

src/validators.ts
```diff
@@ -539,7 +591,12 @@ export const vStorageOptions = v.object({
 });
 
 const vPromptFields = {
-  system: v.optional(v.string()),
+  system: v.optional(
+    v.union(v.string(), vSystemMessage, v.array(vSystemMessage)),
+  ),
+  instructions: v.optional(
+    v.union(v.string(), vSystemMessage, v.array(vSystemMessage)),
+  ),
```
ianmacartney: would it be too annoying to always serialize instructions to `system` so the model only has one field for this, and round-trips it to `instructions` as part of mapping?

ianmacartney (approved):
Directionally it looks good - I assume you will own validation of the new behavior, and specifically around the `previousResponseMessageCount`-related changes. 
If you disagree or want further follow-ups on any comments, you can re-request review. Otherwise I trust you to make necessary changes before merging

ianmacartney:
You can use `>= @convex-dev/rag@0.8.0-alpha.0` as a v7 peer dependency - I haven't verified it but it looks good. Tracking in https://github.com/get-convex/rag/pull/108

src/vercel/client/streamText.ts
robelest: They own separate state: the pending message, component stream, and user callback. Internal cleanup now shares one abort promise, while all three are still attempted.

src/vercel/client/definePlaygroundAPI.ts
robelest: `step.response.messages` is step-local in v7. We now serialize exactly that step, with multi-step count and role-order coverage.

src/validators.ts
robelest: Done. The Convex action wire stores only `system`; the Vercel boundary maps it to `instructions`.

src/vercel/client/start.ts
robelest: This opts the raw handler into request bodies and non-streaming response bodies, which AI SDK 7 omits by default. Explicit caller settings still win.

src/vercel/mapping.ts
robelest: No reverse mapping is needed in v7; warnings are persisted output metadata, not model-message input. The unused reverse mapper is removed.

PR #307, opened 2026-07-30
feat(persistence): persist oversized AI SDK 7 streamed files

robelest:
## Summary

- Persist oversized streamed AI SDK 7 file content through the component's existing storage-reference lifecycle.
- Materialize supported image, reasoning-file, and canonical tool-result file payloads while leaving arbitrary tool output opaque.
- Balance file ownership across stream writes, pending-message replacement, normal finalization, recovery, deletion, and failure cleanup.
- Reuse generic component storage references and refcounts instead of adding provider-specific storage state.
- Preserve file metadata, nested tool-output ownership, deduplication, and cleanup behavior across serialization and recovery.

## Review boundary

This PR adds durable streamed-file persistence on top of the completed AI SDK 7 runtime migration in #306. It does not add another message representation or broaden the runtime migration.

## Compatibility

- Existing persisted AI SDK 6 and AI SDK 7 rows remain readable.
- Schema changes are additive and require no data migration.
- Registration remains ownership-neutral; references and refcounts define durable ownership.

## Validation

Clean install, typecheck, lint, build, and all 348 tests pass on the stack tip.

## Stack

1. #304 — isolate the provider adapter behind a neutral stream codec
2. #305 — extend the codec for the AI SDK 7 chunk superset
3. #306 — migrate the Vercel runtime to AI SDK 7
4. **#307 — persist oversized streamed files**

### DIFF
--- a/src/component/files.ts
+++ b/src/component/files.ts
@@ -1,9 +1,11 @@
 import { paginator } from "convex-helpers/server/pagination";
-import type { Id } from "./_generated/dataModel.js";
+import type { Doc, Id } from "./_generated/dataModel.js";
 import { mutation, type MutationCtx, query } from "./_generated/server.js";
 import { schema, v } from "./schema.js";
 import { paginationOptsValidator } from "convex/server";
 import type { Infer } from "convex/values";
+
+const FILE_CLEANUP_GRACE_MS = 24 * 60 * 60 * 1000;
 
 const addFileArgs = v.object({
   storageId: v.string(),
@@ -28,18 +30,21 @@ export async function addFileHandler(
   args: Infer<typeof addFileArgs>,
 ) {
   // Support both mediaType (preferred) and mimeType (deprecated)
-  const mediaType = args.mediaType ?? args.mimeType;
+  const mediaType = normalizeMediaType(args.mediaType ?? args.mimeType);
 
-  const existingFile = await ctx.db
+  const existingFiles = await ctx.db
     .query("files")
     .withIndex("hash", (q) => q.eq("hash", args.hash))
     // eslint-disable-next-line @convex-dev/no-filter-in-query -- We do not expect many files with the same hash and different filenames
     .filter((q) => q.eq(q.field("filename"), args.filename))
-    .first();
+    .collect();
+  const existingFile = existingFiles.find((file) =>
+    sameMediaType(file, mediaType),
+  );
   if (existingFile) {
-    // increment the refcount
+    // Registration is not ownership. Persisted messages and in-flight streams
+    // acquire ownership explicitly.
     await ctx.db.patch("files", existingFile._id, {
-      refcount: existingFile.refcount + 1,
       lastTouchedAt: Date.now(),
     });
     return {
@@ -83,14 +88,18 @@ export const useExistingFile = mutation({
   args: {
     hash: v.string(),
     filename: v.optional(v.string()),
+    mediaType: v.optional(v.string()),
   },
   handler: async (ctx, args) => {
-    const file = await ctx.db
+    const files = await ctx.db
       .query("files")
       .withIndex("hash", (q) => q.eq("hash", args.hash))
       // eslint-disable-next-line @convex-dev/no-filter-in-query -- We do not expect many files with the same hash and different filenames
       .filter((q) => q.eq(q.field("filename"), args.filename))
-      .first();
+      .collect();
+    const file = files.find((candidate) =>
+      sameMediaType(candidate, normalizeMediaType(args.mediaType)),
+    );
     if (!file) {
       return null;
     }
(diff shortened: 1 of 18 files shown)


src/component/files.ts
```diff
@@ -28,18 +28,21 @@ export async function addFileHandler(
...
+    .collect();
+  const existingFile = existingFiles.find((file) =>
+    sameMediaType(file, mediaType),
+  );
   if (existingFile) {
-    // increment the refcount
+    // Registration is not ownership. Messages and in-flight streams acquire
+    // ownership explicitly when they persist a reference to this file.
     await ctx.db.patch("files", existingFile._id, {
-      refcount: existingFile.refcount + 1,
```
ianmacartney: were we leaking all files before? Or could there be a race now where an existing file temporarily has refcount: 0 before being attributed to a new stream in a separate transaction?

src/component/streams.ts
```diff
@@ -48,7 +56,26 @@ export const addDelta = mutation({
...
+    const { fileRefs, ...delta } = args;
+    if (fileRefs?.length) {
+      const previous = stream.fileRefs ?? [];
+      const aliases = new Map(
+        previous.map((ref) => [`${ref.url}\0${ref.fileId}`, ref] as const),
+      );
+      for (const ref of fileRefs) {
+        aliases.set(`${ref.url}\0${ref.fileId}`, ref);
+      }
+      const next = [...aliases.values()];
```
ianmacartney: this seems a bit funny to me - is it just finding files that have unique url/fileId pairs? Why would the fileId not be unique on its own?

src/component/files.ts
```diff
@@ -28,18 +28,21 @@ export async function addFileHandler(
...
+  const existingFiles = await ctx.db
     .query("files")
     .withIndex("hash", (q) => q.eq("hash", args.hash))
     // eslint-disable-next-line @convex-dev/no-filter-in-query -- We do not expect many files with the same hash and different filenames
     .filter((q) => q.eq(q.field("filename"), args.filename))
-    .first();
+    .collect();
+  const existingFile = existingFiles.find((file) =>
+    sameMediaType(file, mediaType),
+  );
```
ianmacartney, line 43: Is the expectation that the same filename could have different media types? or that something omits the filename and wants to be distinguished by media type? is there an example here to worry about? the same content being served as .txt and .html?

src/streaming/materializePersistedUIMessageChunks.ts
```diff
@@ -212,14 +267,21 @@ export function projectPersistedUIMessageChunkParts(
       }
     }
     if (toolContent.length > 0) {
+      const fileIds = projectFileIds(
+        current,
+        fileRefs,
+        false,
+        (part) => part.providerExecuted !== true,
```
ianmacartney: can you explain this? Not that I think there's a bug, but not sure what the providerExecuted contract is w.r.t. files

src/streaming/materializePersistedUIMessageChunks.ts
```diff
@@ -249,6 +312,56 @@ export function projectPersistedUIMessageChunkParts(
...
+    if (
+      includeTopLevel &&
+      (part.type === "file" || part.type === "reasoning-file")
+    ) {
+      urls.add(part.url);
+    }
+    if (
+      isToolPart(part) &&
+      part.state === "output-available" &&
+      includeTool(part)
```
ianmacartney: this feels pretty AI SDK-specific, despite not being in /vercel/ - I guess a lot of the UIMessageChunk logic is tightly coupled for now. Not the worst, but curious how you're thinking about streaming generically vs. from AI SDK

src/vercel/client/streaming.ts
```diff
@@ -239,32 +250,23 @@ export class DeltaStreamer<T> {
...
-              reason: "abortSignal",
-            });
-          }
-        } catch {
-          // Best-effort cleanup — the stream will be garbage-collected
-          // by the 10-minute timeout if this fails.
-        }
-      });
+      const abortFromSignal = () => {
+        void this.#abort("abortSignal").catch(() => {
```
ianmacartney, line 259: if nothing else awaits the abort, do we potentially lose out here? I noticed we don't await the abort signal from most cleanup places

src/vercel/client/streaming.ts
```diff
@@ -334,75 +338,140 @@ export class DeltaStreamer<T> {
...
+    const drain = this.#drainDeltas();
+    this.#ongoingWrite = drain;
+    void drain.finally(() => {
+      if (this.#ongoingWrite === drain) this.#ongoingWrite = undefined;
+    });
+    return drain;
+  }
+
+  async #drainDeltas() {
+    while (!this.abortController.signal.aborted) {
```
ianmacartney: like should we await the aborted promise after this while loop if it exists? just skimming here, so might be off

ianmacartney (approved):
looking good. Sorry if I messed up the merge conflicts by merging the other PR. I'll let you merge this after responding to feedback, or re-request if you want another pass

src/component/files.ts
robelest: Registration used to increment refcount, so repeated registration without persistence could leak ownership. It now leaves refcount unchanged, and normal cleanup excludes zero-ref files for 24 hours after registration or an ownership change. That covers the gap before a stream or message mutation takes ownership.

src/component/streams.ts
robelest: This preserves the URL-to-file association used when persisted chunks are projected back to Messages. One file can have multiple URLs, so deduping on fileId alone would lose an association. The map is now named refsByUrlAndFileId.

src/component/files.ts
robelest: Filename and media type are observable FileUIPart fields. Identical bytes can represent distinct files for consumers, such as text/plain and text/html, so both are part of the durable identity.

src/streaming/materializePersistedUIMessageChunks.ts
robelest: Provider-executed tool results are stored on the assistant message. Locally executed tool results are stored on the tool message. The projection uses providerExecuted to assign file ownership to the message that contains the result.

src/streaming/materializePersistedUIMessageChunks.ts
robelest: It lives here because recovery has to turn persisted chunks back into Messages, and UIMessageChunk is the only format we write today. I do not think a codec layer buys us much yet.

src/vercel/client/streaming.ts
robelest: consumeStream now joins the shared abort promise when the source throws after an external abort. The regression test holds an in-flight delta write and verifies consumption does not settle before the component abort finishes. The signal listener stays fire-and-forget because event listeners cannot be awaited.

src/vercel/client/streaming.ts
robelest: The join is in consumeStream, outside the delta writer. Waiting inside the writer would create a cycle because signal cleanup already waits for the active write.

src/streaming/materializePersistedUIMessageChunks.ts
ianmacartney: Gotcha, thanks for that context. I forgot that

src/component/streams.ts
ianmacartney: How can a file have multiple URLs? Maybe in a future where you can generate multiple time-bound file storage URLs? Today getUrl is always the same for a given storageId

src/component/files.ts
ianmacartney: Gotcha. I think the safeguard I had missed here is that by bumping the last modified timestamp on refcount change, we can buy ourselves time between a decrement of an old file and attributing it to a new owner. I was worried that old files would be immediately eligible for vacuuming when deallocated

ianmacartney: Oh I think I remember the case I was worried about: A file with refcount 0 has been stale for 23hrs 59mins. In the final moment, a request with the same bytes calls addFile and gets de-duped. But by the time it tries to claim it, it's been cleaned up. 
The safeguard is then to transactionally bump the updated time on a cache hit (especially if the refcount is 0).

src/component/streams.ts
robelest: Recovery resolves ownership from the URL persisted in each chunk, so `fileRefs` keeps the URL-to-file association. A URL now rejects a different file ID; multiple URLs for one file keep a single ownership reference. This also handles storage URLs that rotate.

src/component/files.ts
robelest: Both cache-hit paths now patch `lastTouchedAt` in the same mutation. The regression reuses a zero-ref file at 23:59 and confirms the original cleanup deadline does not make it eligible.

PR #326, opened 2026-08-23
fix(streaming): drain deltas at end-of-stream, not at the final step

### LINKED ISSUE #323: saveStreamDeltas: final step's save races the throttle buffer → tail of fast generations never persisted as deltas
**Version:** `@convex-dev/agent@0.7.0` (same code on `main` today), `ai@7.0.67`, `saveStreamDeltas: { returnImmediately: true, throttleMs: 100 }`.

## What happened

We serve resumable HTTP streams from the persisted deltas (`streams.list` + `streams.listDeltas`, polled). On short generations the delta log ends mid-generation: trailing text, tool-call parts and the finish chunk never arrive, even though the saved message row is complete. How much is missing scales with how fast the generation finishes relative to the writer's flush cadence. Under `convex-test`, where the whole action settles before any throttled flush fires, the delta log ends up empty.

Sequence, as far as I can trace it:

1. The final step's `onStepEnd` calls `streamer.markFinishedExternally()` and saves the message atomically with the stream finish (`src/vercel/client/streamText.ts:216-218`).
2. `DeltaStreamer.addParts` early-returns once `#finishedExternally` is set (`src/vercel/client/streaming.ts:292-311`, "Late deltas would be silently dropped by `streams.addDelta` — skip the work"), and the flushing `finish()` never runs for the buffered remainder.
3. The component refuses deltas on a `finished` stream row, so whatever the throttle buffer held at that moment is gone.

Awaiting the drain instead of using `returnImmediately: true` doesn't help, and lowering `throttleMs` only shrinks the window: in the awaited mode the flag is set before the consumer drains at all.

## Why it matters

For the resumable streaming pattern the delta log is the record of an in-flight generation, and it ends up incomplete. Most consumers never notice because `useUIMessages` replaces the stream with the saved message the moment the save commits, which also hides the bug in testing. Anything that reads the delta log directly, like an HTTP replay endpoint built on `streams.listDeltas`, gets the truncated version every time.

## Expected

Pending deltas flush before, or atomically with, the finishing save. `finishHandler` / `abortById` already take an optional `finalDelta` that is inserted before the state check; the client never passes one on the external-finish path.

## Workaround

At terminal turn status we diff delivered chunks against the persisted message and synthesize the missing suffix (drops are always a suffix), continuing the interrupted text part's id so the tail doesn't render as a separate text block. We'd delete that on a fix.

robelest:
Fixes https://github.com/get-convex/agent/issues/323

`saveStreamDeltas` discarded buffered deltas. The final step's `onStepEnd` marked the stream finished externally, which stopped `addParts` and made `consumeStream` skip `finish()`, the only thing that drains the buffer. A short generation could persist a single `start` part while the saved message row was complete.

The drain has to happen at EOF, not at `onStepEnd`. Simpler fixes there each fail at a different point:

```
drain once before setting the flag
  └─ misses writes #sendDelta reassigns from its own tail
     └─ loop with an identity check
        └─ misses a part parked in getStreamId()
           └─ buffer before that await
              └─ misses a part still inside consumeStream's next()
```

Parts reach us through a teed UI message stream and sit in the SDK's transform pipeline until the iterator yields them, and the stream-level `finish` chunk is emitted strictly after the last `onStepEnd`. So no earlier callback can see everything that is still coming.

**What changed**

- `#finishedExternally` split in two: `#finishHandledExternally` (who calls `streams.finish`, fixed at construction) and `#stoppedAccepting` (set only where the row is finished before the source is drained).
- `consumeStream` drains at EOF and finishes only when ownership is internal.
- The awaited path stops touching the streamer in `onStepEnd`. It records `pendingFinalStep` and lets the row keep accepting, so it now persists the trailing chunks it used to drop, `finish` included. Race free by sequencing: pull, pull, EOF, flush, save and finish.
- `returnImmediately` still flushes at `onStepEnd`, since nothing awaits consumption there and #265 needs the inline save. Parts still inside the SDK at that instant never become deltas. Documented on `flushAndStopAccepting` rather than papered over.

One new test covers the contract that an out of band abort fails the generation rather than saving a successful message onto an aborted row.


ianmacartney (approved):
sounds reasonable. I'm a bit wary of all the state being encoded of the form "don't worry someone else is responsible for doing XYZ" but I don't have a specific alternative in mind. If we can find a good layering over time such that each piece is hard to use wrong / has scoped lifetime by default, that would be awesome. Maybe this all gets easier in a world without driving everything from ai sdk callbacks?

PR #368, opened 2026-09-14
Add support for Convex gateway in example app

davidjwiner:
Add the Convex gateway to the example app instead of using the mock language model. 

I think the mock is still used in tests so I left it in for now.

Before:

CleanShot 2026-09-14 at 4 00 51 PM

After:

CleanShot 2026-09-14 at 4 01 12 PM

----


ianmacartney (approved):
I think the coderabbit comment is my only concern, but I think generally folks running the example here have forked the component and maybe we can assume those users are already paid customers? I wonder if we could have a fallback somehow - if generation fails b/c of AiGatewayDisabled/Unavailable, printing the warning to set the env var. 
I'll probably only hit this when testing local deployments

davidjwiner:
> I think the coderabbit comment is my only concern, but I think generally folks running the example here have forked the component and maybe we can assume those users are already paid customers? I wonder if we could have a fallback somehow - if generation fails b/c of AiGatewayDisabled/Unavailable, printing the warning to set the env var. I'll probably only hit this when testing local deployments

Yeah that's less than ideal for the local / free plan cases... I took a look at the docs and tried to make it so if the AI gateway is disabled we still fallback to the mock

davidjwiner:
@coderabbitai re-review

example/convex/modelsForDemo.ts
```diff
@@ -20,7 +22,48 @@ if (process.env.ANTHROPIC_API_KEY) {
...
+}
+
+let gatewayAvailable: boolean | undefined;
+
+async function gatewayIsAvailable(): Promise<boolean> {
+  if (gatewayAvailable !== undefined) return gatewayAvailable;
+  try {
+    await getServiceToken("ai-gateway");
+    gatewayAvailable = true;
+  } catch (error) {
```
ianmacartney, line 37: I think this will be called at import time. I'm not sure our support for this syscall at import time. did you get a chance to try this with a local backend?

davidjwiner: When running with a local backend it seems to work (doesn't crash). The dummy text appears in the chat window and there's a warning that appears in the console:

```
9/14/2026, 6:52:15 PM [CONVEX A(chat/basic:generateResponse)] [WARN] 'The AI gateway can\'t be used from this deployment. Examples will respond with mock text. Run "npx convex env set GROQ_API_KEY=<your-api-key>" or"npx convex env set OPENAI_API_KEY=<your-api-key>" from the example directory to use a real model.\n\nSyscall error: Error: `getServiceToken("ai-gateway")` isn\'t available on this deployment because the AI gateway is a Convex Cloud service. Deploy to Convex Cloud, or call your model provider directly with your own API key.
```

I couldn't figure out a great way to figure out at runtime if we're running with a local backend. Claude suggested checking process.env.CONVEX_CLOUD_URL — I can do that if you think it's better than this solution!

ianmacartney: `ctx.meta.getDeploymentMetadata()` currently returns `null` for region in these cases... which is fine I guess. I have a PR adding the deployment "type" (prod/dev/preview/custom/local), but that is on the back burner

ianmacartney: but that would require having `ctx` available, which isn't the case at import time. just FYI localhost might look different on windows vs. mac iirc

